PR #13, opened 2026-03-09
Add Save To Project node

### LINKED ISSUE #11: Project Saver Node

zachgiordano:
Add Save To Project node

Closes #11

### DIFF
--- a/griptape_nodes_library/files/save_to_project.py
+++ b/griptape_nodes_library/files/save_to_project.py
@@ -1,7 +1,7 @@
 """SaveToProject node - save a file to the project using situation-based path resolution."""
 
 import logging
-from pathlib import Path, PurePosixPath
+from pathlib import Path, PureWindowsPath
 from typing import Any
 
 from griptape.artifacts import UrlArtifact
@@ -92,7 +92,7 @@ class SaveToProject(SuccessFailureNode):
 
         return exceptions or None
 
-    def process(self) -> None:
+    async def aprocess(self) -> None:
         self._clear_execution_status()
 
         source_value = self.get_parameter_value("source")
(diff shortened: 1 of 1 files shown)


griptape_nodes_library/files/save_to_project.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+
+    def _update_default_filename(self, source_path: str) -> None:
+        """Update the default output filename extension based on the source path."""
+        output_file_value = self.get_parameter_value("output_file")
+        if output_file_value and output_file_value != "output":
+            return
+
+        suffix = PurePosixPath(source_path).suffix or Path(source_path).suffix
+        new_default = f"output{suffix}" if suffix else "output"
```
SavagePencil: suffix comes in with the leading dot, I'm guessing

griptape_nodes_library/files/save_to_project.py
```diff
@@ -0,0 +1,150 @@
...
+            result_details=f"File saved successfully to {saved_path}",
+        )
+
+    def _update_default_filename(self, source_path: str) -> None:
+        """Update the default output filename extension based on the source path."""
+        output_file_value = self.get_parameter_value("output_file")
+        if output_file_value and output_file_value != "output":
+            return
+
+        suffix = PurePosixPath(source_path).suffix or Path(source_path).suffix
```
SavagePencil: Not familiar with PurePosixPath, why do we need to do this OR here?

FUQ: will this play nice with network drive paths (e.g., `\\Shared\Finals\blah.png`)?

griptape_nodes_library/files/save_to_project.py
zachgiordano: Yes, that's correct. PurePosixPath.suffix and Path.suffix both return the extension with the leading dot (e.g., .png). So f"output{suffix}" produces output.png.

griptape_nodes_library/files/save_to_project.py
zachgiordano: Good call, I double checked the usage. PureWindowsPath handles both forward slashes and backslashes as separators, so that's actually the class we should use to handle paths from any platform. Will update and include a comment.

griptape_nodes_library/files/save_to_project.py
```diff
@@ -0,0 +1,150 @@
...
+    if isinstance(value, UrlArtifact):
+        return str(value.value)
+    if isinstance(value, str):
+        return value
+    if isinstance(value, dict) and "value" in value:
+        return str(value["value"])
+    return None
+
+
+class SaveToProject(SuccessFailureNode):
```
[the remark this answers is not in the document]
SavagePencil: oh duh, yes.

PR #98, opened 2026-03-25
Add Publish to Folder publishing target to the standard library

### LINKED ISSUE #94: Publish to Folder
Publish a workflow and it's dependencies to a directory local to the engine instance.

Similar to but simpler than the Nuke Library: https://github.com/griptape-ai/griptape-nodes-library-nuke

zachgiordano:
* Add Publish to Folder publishing target to the standard library

Publishes contents like this:

published_workflow_package

Closes #94

### DIFF
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -1,6 +1,7 @@
 {
   "name": "Griptape Nodes Library",
   "library_schema_version": "0.6.0",
+  "advanced_library_path": "griptape_nodes_library/griptape_nodes_library_advanced.py",
   "settings": [
     {
       "description": "API keys required by nodes in this library",
--- a/griptape_nodes_library/publish_workflow/local_publish_options.py
+++ b/griptape_nodes_library/publish_workflow/local_publish_options.py
@@ -1,0 +1,73 @@
+"""Publish-time options provider for the local folder publisher.
+
+Called by the engine when the frontend opens the publish dialog. Returns a
+single field for the destination directory. If a previous publish saved a
+config under the StartFlow node's metadata, that value pre-populates the
+field.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+from griptape_nodes.retained_mode.events.config_events import (
+    GetConfigValueRequest,
+    GetConfigValueResultSuccess,
+)
+from griptape_nodes.retained_mode.events.flow_events import GetTopLevelFlowRequest, GetTopLevelFlowResultSuccess
+from griptape_nodes.retained_mode.events.workflow_events import (
+    GetPublishOptionsRequest,
+    GetPublishOptionsResultSuccess,
+    PublishFieldType,
+    PublishOptionField,
+)
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+PUBLISH_OUTPUT_DIRECTORY_FIELD = "publish_output_directory"
+
+
+def _find_start_flow_node():  # noqa: ANN202
+    result = GriptapeNodes.handle_request(GetTopLevelFlowRequest())
+    if not isinstance(result, GetTopLevelFlowResultSuccess) or result.flow_name is None:
+        return None
+    control_flow = GriptapeNodes.FlowManager().get_flow_by_name(result.flow_name)
+    for node in control_flow.nodes.values():
+        if node.__class__.__name__ == "StartFlow":
+            return node
+    return None
+
+
+def _get_workspace_directory() -> str:
+    result = GriptapeNodes.handle_request(GetConfigValueRequest(category_and_key="workspace_directory"))
+    if isinstance(result, GetConfigValueResultSuccess) and result.value:
+        return str(result.value)
+    return str(Path.home())
+
+
+def get_local_publish_options(request: GetPublishOptionsRequest) -> GetPublishOptionsResultSuccess:
+    """Build the list of fields for the local folder publish dialog."""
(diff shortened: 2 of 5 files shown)


griptape_nodes_library/workflows/start_flow.py
```diff
@@ -1,9 +1,13 @@
 from typing import Any
 
+from griptape_nodes.exe_types.core_types import Parameter, ParameterGroup, ParameterMode
 from griptape_nodes.exe_types.node_types import StartNode
+from griptape_nodes.traits.file_system_picker import FileSystemPicker
 
 
 class StartFlow(StartNode):
```
zachgiordano, line 6: Updating the StartFlow node to include publishing configuration for this Publish To Folder functionality may no longer be required. Publishing options can be registered by the Library instead, like what the Nuke library does: https://github.com/griptape-ai/griptape-nodes-library-nuke/blob/main/nuke_nodes/nuke_library_advanced.py#L38-L50

The example publish options are defined here: https://github.com/griptape-ai/griptape-nodes-library-nuke/blob/main/publish_gizmo/nuke_publish_options.py

griptape_nodes_library/publish_workflow/local_publisher.py
```diff
@@ -0,0 +1,685 @@
...
+            )
+        except Exception as e:
+            details = f"Failed to publish workflow '{self._workflow_name}'. Error: {e}"
+            logger.error(details)
+            return PublishWorkflowResultFailure(result_details=details, exception=e)
+
+    # -- Destination path --
+
+    def _get_destination_path(self) -> str | None:
+        """Read the publish_output_directory from the StartFlow node."""
```
zachgiordano: If we go with the publish options approach, then we no longer need to inspect the nodes of the workflow to grab the StartFlow node and try to get the `publish_output_directory` Parameter's value.

https://github.com/griptape-ai/griptape-nodes/blob/main/src/griptape_nodes/retained_mode/events/workflow_events.py#L467

zachgiordano (commented):
This PR looks good, but is stale and should be rebased against current `main`. Additionally, some publishing Event improvements have  been made which could clean this up.

zachgiordano:
@griptapeops Can you update this PR to address the comments?

griptape_nodes_library/publish_workflow/local_publish_options.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def get_local_publish_options(request: GetPublishOptionsRequest) -> GetPublishOptionsResultSuccess:
+    """Build the list of fields for the local folder publish dialog."""
+    current: dict = request.current_selections or {}
+
+    if not current:
+        start_flow = _find_start_flow_node()
+        if start_flow is not None:
+            saved = start_flow.metadata.get("publish_config")
```
kateforsberg13, line 54: should the publish options be on the start flow node or in the library itself? did you make them on start flow node so the standard library could have other types of publishing?

griptape_nodes_library.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "Griptape Nodes Library",
   "library_schema_version": "0.6.0",
+  "advanced_library_path": "griptape_nodes_library/griptape_nodes_library_advanced.py",
```
kateforsberg13, line 4: should this pr just be in the gtn standard library repo? since there aren't engine changes? as well, should be increment the library version, and make sure it's dependent on the engine version that has a workflowpackager.

griptape_nodes_library/publish_workflow/local_publisher.py
```diff
@@ -0,0 +1,178 @@
...
+
+            workflow = WorkflowRegistry.get_workflow_by_name(self._workflow_name)
+            destination = Path(str(destination_path).strip()) / self._workflow_name
+
+            self._packager.emit_progress(5.0, "Preparing output directory...")
+            destination.mkdir(parents=True, exist_ok=True)
+
+            # Bundle workflow file, libraries, config, .env, project template, static files,
+            # HuggingFace download script, and pyproject.toml.
+            self._packager.package_to_folder(destination, workflow)
```
kateforsberg13, line 61: nice, glad we are using this since it's in the engine

kateforsberg13 (changes requested):

griptape_nodes_library/publish_workflow/local_publish_options.py
zachgiordano: The publish options are registered by the Library, yes. This is the fallback method of trying to detect previously set publish options stored on the StartFlow node's metadata, like what the Nuke library does: https://github.com/griptape-ai/griptape-nodes-library-nuke/blob/main/publish_gizmo/nuke_publish_options.py#L49-L56

griptape_nodes_library.json
zachgiordano: > should this pr just be in the gtn standard library repo

I'm not sure I follow. This PR currently only impacts the `griptape-nodes-library-standard` repo. All required engine changes are already in place.

> make sure it's dependent on the engine version that has a workflowpackager

Yes, good call out. This is covered already with the engine version requirement of `0.82.0`. The [WorkflowPackager](https://github.com/griptape-ai/griptape-nodes/compare/v0.80.2...v0.81.2) should've been released somewhere in v0.81.x.

kateforsberg13: sorry I was so confused. It isn't in the engine, you're right.

kateforsberg13 (approved):

PR #114, opened 2026-03-30
Create Workflow Node

kateforsberg13:
This node runs a locally saved workflow, within another locally saved workflow. 
There is special handling for this node in the gui, and in the engine. 
Engine PR will have to be merged first, then vsl-gui, then this one.

### DIFF
--- a/griptape_nodes_library/engine/subflow_workflow_node.py
+++ b/griptape_nodes_library/engine/subflow_workflow_node.py
@@ -16,13 +16,15 @@ from griptape_nodes.retained_mode.events.parameter_events import (
     RemoveParameterFromNodeRequest,
     SetParameterValueRequest,
 )
+
+# Add pyright ignore because this is a new event - we know it does exist
 from griptape_nodes.retained_mode.events.workflow_events import (
     ImportWorkflowAsReferencedSubFlowRequest,
     ImportWorkflowAsReferencedSubFlowResultSuccess,
-    ListCallableWorkflowsRequest,
-    ListCallableWorkflowsResultSuccess,
+    ListCallableWorkflowsRequest,  # pyright: ignore[reportAttributeAccessIssue]
+    ListCallableWorkflowsResultSuccess,  # pyright: ignore[reportAttributeAccessIssue]
 )
-from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes, logger
 from griptape_nodes.traits.button import Button, ButtonDetailsMessagePayload
 from griptape_nodes.traits.options import Options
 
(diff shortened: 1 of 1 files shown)


griptape_nodes_library.json
```diff
@@ -2682,6 +2681,17 @@
         "icon": "FolderCog"
       }
     },
+    {
```
collindutter: Accidental commit?

griptape_nodes_library/engine/subflow_workflow_node.py
```diff
@@ -0,0 +1,274 @@
...
+
+        if not workflow_name or not WorkflowRegistry.has_workflow_with_name(workflow_name):
+            return
+
+        workflow = WorkflowRegistry.get_workflow_by_name(workflow_name)
+        file_path = WorkflowRegistry.get_complete_file_path(workflow.file_path)
+        content = Path(file_path).read_text(encoding="utf-8")
+
+        existing_flows = set(GriptapeNodes.ObjectManager().get_filtered_subset(type=ControlFlow).keys())
+        exec(content, {"__file__": file_path, "__name__": "workflow_module", "__builtins__": __builtins__})  # noqa: S102
```
collindutter: Should the node be `exec`ing? This seems like something we should try to hand off to the engine.

collindutter (changes requested):

griptape_nodes_library/engine/subflow_workflow_node.py
```diff
@@ -0,0 +1,274 @@
...
+                    continue
+                if param_name in end_node.parameter_output_values:
+                    value = end_node.parameter_output_values[param_name]
+                else:
+                    value = end_node.get_parameter_value(param_name)
+                if value is not None:
+                    self.parameter_output_values[param_name] = value
+
+    def process(self) -> Any:
+        raise NotImplementedError
```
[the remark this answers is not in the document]
kateforsberg13: yes, I only implemented aprocess, but if it calls process then the node won't run, so I wanted it to fail. should I just not override process at all?

griptape_nodes_library.json
cjkindel: This PR will wait on a new engine to get cut, right (so the library can require that engine version)?

griptape_nodes_library/engine/subflow_workflow_node.py
```diff
@@ -0,0 +1,276 @@
...
+        )
+        self.add_parameter(self.workflow_file)
+        self.metadata["workflow_node"] = True
+
+        # Pre-populate dynamic parameters so they exist before connections and
+        # set-value commands are applied during deserialization.
+        if saved_workflow and saved_workflow in workflow_names:
+            # Ensure workflow_shape_params is empty so _remove_workflow_shape_parameters
+            # is a no-op (the node isn't registered yet, so the request would fail).
+            self.metadata["workflow_shape_params"] = []
```
SavagePencil: presumably we can get rid of all this metadata stuff once I get you a TransitionPlan component; we will want to validate all params match what the workflow currently has in it when this node gets instantiated.

griptape_nodes_library/engine/subflow_workflow_node.py
```diff
@@ -0,0 +1,287 @@
...
+                self._reload_subflow(value)
+        super().after_value_set(parameter, value)
+
+    def get_node_dependencies(self) -> NodeDependencies | None:
+        deps = super().get_node_dependencies()
+        if deps is None:
+            deps = NodeDependencies()
+        workflow_name = self.get_parameter_value("workflow_file")
+        if workflow_name and WorkflowRegistry.has_workflow_with_name(workflow_name):
+            deps.referenced_workflows.add(workflow_name)
```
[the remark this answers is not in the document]
SavagePencil: YEah, let's start with warning. We can attack this once it's a real problem.

PR #122, opened 2026-04-03
fix: eagerly register variables in CreateVariable nodes

collindutter:
Fixes #121.

`CreateVariable` previously only registered variables with the `VariablesManager` during `process()`, which only runs as part of a workflow execution. This meant other nodes like `GetVariable` couldn't reference the variable until `CreateVariable` had actually run, breaking the expected user experience.

This adds eager variable registration in `before_value_set` on the `variable_name` parameter. When the user enters a variable name, the variable is immediately created in the `VariablesManager`. When the variable name is changed, the old variable is removed and a new one is created. When the variable name is cleared, the old variable is removed.

The `process()` method continues to handle the full create-or-update logic during workflow runs, ensuring values and types are kept in sync.


shhlife (changes requested):
this is very cool that it will create the variable when you type in the name - but it'd be awesome if it also would update the variable type when you specify it.

For example - here I created a variable and ran the Engine node to see what variables existed.. it shows me the variable, but the type is `''`.. which is fine, I haven't set it yet.

Image

However, if I set the type, it still shows up as `''`:

Image

If I execute the Create Variable node, then the type shows up correctly.

Image

also, as long as we're updating this - can we also have get_variable provide a dropdown list of existing variables? We could allow for custom & enter the name if we want - but it'd be great to have an existing list for users to choose.

This should be able to be grabbed using the same `ListVariableRequest` event.

PR #135, opened 2026-04-09
World Labs World Generation Node

andrewfrench:
At long last, the World Labs World Generation node supporting gaussian splat scene generation and BYOK support.

### DIFF
--- a/griptape_nodes_library/splat/splat_artifact.py
+++ b/griptape_nodes_library/splat/splat_artifact.py
@@ -1,0 +1,19 @@
+from typing import Any
+
+from griptape.artifacts.url_artifact import UrlArtifact
+
+
+class SplatUrlArtifact(UrlArtifact):
+    """A splat file URL artifact."""
+
+    def __init__(self, value: str, meta: dict[str, Any] | None = None) -> None:
+        super().__init__(value=value, meta=meta or {})
+
+    def to_text(self) -> str:
+        """Convert the splat URL to text representation."""
+        return f"Splat file URL: {self.value}"
+
+    def to_dict(self) -> dict[str, Any]:
+        data = super().to_dict()
+        data["metadata"] = data.get("meta", {})
+        return data
--- a/griptape_nodes_library/splat/world_labs_world_generation.py
+++ b/griptape_nodes_library/splat/world_labs_world_generation.py
@@ -1,0 +1,797 @@
+from __future__ import annotations
+
+import asyncio
+import base64
+import logging
+from io import BytesIO
+from typing import Any
+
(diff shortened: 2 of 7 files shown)


griptape_nodes_library/splat/world_labs_world_generation.py
```diff
@@ -0,0 +1,797 @@
...
+        )
+
+        # Seed parameter
+        self._seed_parameter = SeedParameter(self, 4294967295)
+        self._seed_parameter.add_input_parameters()
+
+        # OUTPUTS
+        self.add_parameter(
+            ParameterString(
+                name="generation_id",
```
cjkindel, line 240: Where do we actually set this?

griptape_nodes_library/splat/world_labs_world_generation.py
```diff
@@ -0,0 +1,797 @@
...
+                format_map = {
+                    "JPEG": "jpg",
+                    "JPG": "jpg",
+                    "PNG": "png",
+                    "WEBP": "webp",
+                }
+                return format_map.get(image_format, "jpg")
+        except Exception:
+            # If not an image, check video magic bytes
+            if media_bytes[:4] == b"ftyp" or media_bytes[4:8] == b"ftyp":
```
cjkindel: Claude says this magic bytes check is wrong?

```
b"ftyp" in MP4 appears at offset 4-7, not 0-3. The first condition can never be true for a valid MP4. The second condition is correct -- the first should be removed.
```

griptape_nodes_library/splat/splat_artifact.py
```diff
@@ -0,0 +1,36 @@
+from typing import Any
+
+from griptape.artifacts import BaseArtifact
+from griptape.artifacts.url_artifact import UrlArtifact
+
+
+class SplatArtifact(BaseArtifact):
```
cjkindel: Do we need SplatArtifact or just SplatUrlArtifact?

griptape_nodes_library/splat/world_labs_world_generation.py
andrewfrench: we set the value in the base proxy node

griptape_nodes_library/splat/splat_artifact.py
andrewfrench: makes sense just to keep SplatUrlArtifact, will remove

griptape_nodes_library/splat/world_labs_world_generation.py
andrewfrench: updated

cjkindel (approved):

PR #139, opened 2026-04-10
feat: add RunwayML video and image generation nodes

collindutter:
Adds `RunwayMLVideoGeneration` and `RunwayMLImageGeneration` proxy nodes for generating video and images through the Griptape Cloud model proxy.

The video node supports three models: Gen-4.5 (text-to-video and image-to-video), Gen-4 Turbo (image-to-video), and Gen-3 Alpha Turbo (image-to-video). It automatically selects the `text_to_video` or `image_to_video` endpoint based on whether an input image is provided. Ratio options update dynamically when the model changes.

The image node supports Gen-4 Image and Gen-4 Image Turbo models via the `text_to_image` endpoint. It requires a reference image, which is converted to a data URI before being sent to the API. Both nodes use data URIs for media inputs rather than `PublicArtifactUrlParameter` because RunwayML rejects URLs that have an `application/octet-stream` Content-Type header.

Depends on the proxy client PR: https://github.com/griptape-ai/griptape-cloud/pull/1920

## Sources

- https://docs.dev.runwayml.com/api
- RunwayML rejects publicly-hosted URLs with incorrect Content-Type headers, so all image inputs are converted to data URIs inline

## Testing with the engine

To test the node end-to-end in the Griptape Nodes UI:

1. Check out both branches:
   - This repo: `feat/runwayml-proxy-node`
   - griptape-cloud: the proxy client branch (see linked PR above)
2. Start the local griptape-cloud: `cd griptape-cloud && make up/debug`
3. Create DB records for the model config (see proxy client PR for setup steps)
4. Start the engine with proxy overrides:
   ```bash
   GT_CLOUD_PROXY_BASE_URL=http://localhost:8000 GT_CLOUD_PROXY_API_KEY=local make run/watch
   ```
5. In the Nodes UI, add a `RunwayML Video Generation` or `RunwayML Image Generation` node and connect it to a workflow
6. Set the prompt and any other parameters, then run the workflow
7. Verify the output video/image appears in the node's output

## Integration test

To run the automated integration tests:

```bash
GT_CLOUD_PROXY_BASE_URL=http://localhost:8000 GT_CLOUD_PROXY_API_KEY=local GT_CLOUD_API_KEY=fake \
  uv run python -m pytest tests/workflows/test_integration_workflows.py -v -k "runway_ml"
```

### DIFF
--- a/.claude/skills/impl-proxy-node/SKILL.md
+++ b/.claude/skills/impl-proxy-node/SKILL.md
@@ -47,6 +47,7 @@ Based on the spec's media type, read the closest existing node:
 - **Audio**: `griptape_nodes_library/audio/eleven_labs_music_generation.py` (simple, raw bytes result)
 - **Image**: `griptape_nodes_library/image/flux_2_image_generation.py` (model mapping, URL-based result, image download)
 - **Video**: `griptape_nodes_library/video/kling_text_to_video_generation.py` (async, video result)
+- **Image/video input handling**: `griptape_nodes_library/video/kling_image_to_video_generation.py` (data URI conversion pattern with `_prepare_image_data_url_async` and `_coerce_image_url_or_data_uri`)
 
 Study the patterns for:
 - Parameter definition in `__init__`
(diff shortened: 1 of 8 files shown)


shhlife:
hey @collindutter any chance we can get this in?

PR #147, opened 2026-04-14
Execute Python clearer documentation and allow assignment to result variable

SmoolPerson:
Addresses this issue: https://github.com/griptape-ai/griptape-nodes-library-standard/issues/119


SmoolPerson:
Should only be merged after https://github.com/griptape-ai/griptape-nodes/pull/4351 is merged

SavagePencil (approved):
Guessing we can make a more advanced node that allows user to specify multiple variables (maybe takes a list of strings) and outputs them as a dict? But this keeps things backwards compatible, which is great.

PR #153, opened 2026-04-17
Static Files as node dependencies

kateforsberg13:
Bundles files as node dependencies when bundling a workflow file.

### DIFF
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -28,7 +28,7 @@
   "metadata": {
     "author": "Griptape, Inc",
     "description": "Default nodes for Griptape Nodes",
-    "library_version": "0.69.2",
+    "library_version": "0.71.0",
     "engine_version": "0.79.0",
     "tags": [
       "Griptape",
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -28,7 +28,7 @@
   "metadata": {
     "author": "Griptape, Inc",
     "description": "Default nodes for Griptape Nodes",
-    "library_version": "0.70.0",
+    "library_version": "0.69.2",
     "engine_version": "0.79.0",
     "tags": [
       "Griptape",
--- a/griptape_nodes_library/audio/load_audio.py
+++ b/griptape_nodes_library/audio/load_audio.py
@@ -2,7 +2,7 @@ from typing import Any
 
 from griptape.artifacts.audio_url_artifact import AudioUrlArtifact
 from griptape_nodes.exe_types.core_types import NodeMessageResult, Parameter
-from griptape_nodes.exe_types.node_types import BaseNode, DataNode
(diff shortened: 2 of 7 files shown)


griptape_nodes_library/audio/load_audio.py
```diff
@@ -73,6 +74,29 @@ def __init__(self, **kwargs) -> None:
...
+    def get_node_dependencies(self) -> NodeDependencies | None:
+        deps = super().get_node_dependencies()
+        if deps is None:
+            deps = NodeDependencies()
+        value = self.get_parameter_value("path")
+        if not value or not isinstance(value, str):
+            return deps
+
+        # Resolve localhost URLs to filesystem paths the packager can handle
+        resolved = resolve_localhost_url_to_path(value)
```
SavagePencil: Hang on, do we need these?

griptape_nodes_library.json
```diff
@@ -28,7 +28,7 @@
   "metadata": {
     "author": "Griptape, Inc",
     "description": "Default nodes for Griptape Nodes",
-    "library_version": "0.70.0",
+    "library_version": "0.69.2",
```
SavagePencil: I think we may need 0.71

SavagePencil (changes requested):

griptape_nodes_library/audio/load_audio.py
kateforsberg13: They for some reason are happening with Load Images - I dont know why? I'm updated on the most recent version

SavagePencil: Are the files outside the project? If so, that’s where we want to tell them it won’t work for portability. @collindutter for input

SavagePencil (approved):

griptape_nodes_library.json
```diff
@@ -28,7 +28,7 @@
   "metadata": {
     "author": "Griptape, Inc",
     "description": "Default nodes for Griptape Nodes",
-    "library_version": "0.70.0",
+    "library_version": "0.71.0",
```
collindutter, line 31: This should be reverted. Library version is no longer bumped in PRs.

griptape_nodes_library/audio/load_audio.py
```diff
@@ -73,6 +73,17 @@ def __init__(self, **kwargs) -> None:
...
+    def get_node_dependencies(self) -> NodeDependencies | None:
+        deps = super().get_node_dependencies()
+        if deps is None:
+            deps = NodeDependencies()
+        value = self.get_parameter_value("path")
+        if not value or not isinstance(value, str):
+            return deps
+
+        deps.static_files.add(value)
+        return deps
```
collindutter, line 85: This is shared across all load nodes, can you please extract to a shared util?

PR #155, opened 2026-04-20
Fix Kling and SeeDream input parameters

andrewfrench:
Addresses a regression in the SeeDream image generation node and Kling image-to-video generation node in which input parameters were not appropriately routed to the request payload.


collindutter (approved):
TIL "elide"

PR #161, opened 2026-04-23
adds gpt-image tag to OpenAiImageGeneration

shhlife:
adds a tag to OpenAiImageGeneration node with `gpt-image`


cjkindel (commented):
Should we add 2? Seems like its better to have the tag over-verbose vs under

andrewfrench (approved):

shhlife:
> Should we add 2? Seems like its better to have the tag over-verbose vs under

@cjkindel I added specific models

cjkindel (approved):

collindutter (approved):

PR #171, opened 2026-04-24
Variable Nodes Overhaul

### LINKED ISSUE #4411: Do not require an explicit Create Variable node; Set variable will create as needed

zachgiordano:
## Variable Nodes Overhaul
* Enables SetVariable Node to create a Variable if it doesn't already exist
* Declare NodeDependencies for VariableReference

Closes https://github.com/griptape-ai/griptape-nodes/issues/4411

Depends on engine release of: https://github.com/griptape-ai/griptape-nodes/pull/4414

### DIFF
--- a/griptape_nodes_library/variables/create_variable.py
+++ b/griptape_nodes_library/variables/create_variable.py
@@ -2,7 +2,14 @@ import logging
 from typing import Any
 
 from griptape_nodes.exe_types.core_types import Parameter, ParameterMode, ParameterTypeBuiltin
-from griptape_nodes.exe_types.node_types import BaseNode, ControlNode
+from griptape_nodes.exe_types.node_types import (
+    BaseNode,
+    ControlNode,
+    NodeDependencies,
+    VariableAccess,
+    VariableReference,
+)
+from griptape_nodes.retained_mode.variable_types import VariableScope
 
 logger = logging.getLogger("griptape_nodes")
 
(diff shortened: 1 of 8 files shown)


griptape_nodes_library/variables/create_variable.py
```diff
@@ -318,6 +325,36 @@ def process(self) -> None:
         self.parameter_output_values["variable_type"] = variable_type
         self.parameter_output_values["value"] = value
 
+    def get_node_dependencies(self) -> NodeDependencies | None:
+        """Declare the variable this node creates or updates so it survives serialization.
+
+        Access is READ_WRITE: ``process()`` calls ``HasVariableRequest`` + ``GetVariableDetailsRequest``
```
SavagePencil, line 331: Hmmmm this is a tough one. But OK.

griptape_nodes_library/variables/get_variable.py
SavagePencil, line 113: can we get rid of all this hacky volatile stuff now?

griptape_nodes_library/variables/set_variable.py
```diff
@@ -39,41 +69,222 @@ def __init__(
...
+        if not has_variable(node_name=self.name, variable_name=old_name, scope=VariableScope.CURRENT_FLOW_ONLY):
+            self._eager_create_variable(new_name)
+            return
+
+        rename_request = RenameVariableRequest(
+            name=old_name,
+            new_name=new_name,
+            lookup_scope=VariableScope.CURRENT_FLOW_ONLY,
+            starting_flow=current_flow_name,
+        )
```
SavagePencil: this feels possibly problematic? Tell me if it handles this case

Set Variable Node A sets variable `X` to `123`.
Set Variable Node B sets variable `Y` to `456`
Set Variable Node C sets variable `X` to `789`

I change Node C from setting variable `X` to setting variable `Y`

Does this in any way break Node A or Node B? Perhaps this is covered.

griptape_nodes_library/variables/set_variable.py
```diff
@@ -39,41 +69,222 @@ def __init__(
...
+                msg = f"Failed to create variable '{variable_name}': {create_result.result_details}"
+                raise TypeError(msg)
 
-        # Set output values.
         self.parameter_output_values[self.variable_name_param.name] = variable_name
 
+    def get_node_dependencies(self) -> NodeDependencies | None:
+        """Declare the variable this node reads/writes so it survives serialization.
+
+        Access is READ_WRITE: ``process()`` calls ``HasVariableRequest`` before deciding whether to
```
SavagePencil, line 240: dang, we DO have to make Set Variable both a read AND a write. Was hoping it could just be a write.

griptape_nodes_library/variables/set_variable.py
zachgiordano: Good catch, confirmed the bug. 

Walking your scenario in the old code: 
A owns X, B owns Y, C silently adopts X (via `_eager_create_variable`'s collision check). Re-pointing C from X → Y is caught by the engine's collision rejection (safe by luck). But re-point C from X → Z (brand-new name) and the eager rename succeeds, moving engine-side X to Z. Node A's variable_name is still "X", so A's next resolve fails and it silently creates a fresh X. That's the orphaning you were pointing at.

Fix: dropped the eager rename path entirely. `before_value_set` now only eagerly creates-if-missing. Editing variable_name is always treated as "point this node at a different variable". There's no UI signal to distinguish that from a rename, and treating every edit as a repoint avoids the aliasing hazard. Any variable left behind mid-session is filtered out at save time by the NodeDependencies-driven serialization from the engine PR. Added a test mirroring your exact scenario (A→X, B→Y, C→X→Z, assert A still sees X).

SavagePencil: I should have taken a job as a compiler

griptape_nodes_library/variables/get_variable.py
zachgiordano: I don't think so, unfortunately, if we want this behavior still. The engine serialization PR was addressing purely save/load. It didn't introduce any dependency signal between SetVariable's writes and GetVariable's reads.

Although, if we no longer want this Node to be able to switch to a dirty state based on a Variable's value changing, then we could remove this override.

What the serialization PR did fix was:

1. SetVariable Node creates variable `X`
2. GetVariable Node can read variable `X`
3. SetVariable Node gets deleted, workflow saved and reloaded
4. GetVariable can still read variable `X` (which still exists in the Workflow)

SavagePencil: 'k I'm fine with it.

SavagePencil: maybe this actually...worked...all along and what you added here makes it consistent.

SavagePencil (approved):

PR #172, opened 2026-04-28
Feat/video frame extractor

alex-rumsey-foundry:
# Summary
Screenshot 2026-04-30 at 12 55 32

- Introduces BaseVideoInputNode, a format-agnostic abstract base for nodes that consume video but produce non-video output (images, audio, metadata). BaseVideoProcessor now inherits from it, so all existing nodes are unaffected.
- Adds ExtractFrames, a node that accepts a list of 1-based frame numbers and extracts each as an image file using ffmpeg.
- Adds VideoPlayerFrameSelector, a custom JS widget for frame-by-frame scrubbing and selection inside the node editor.

# Test plan
- Instantiate an existing BaseVideoProcessor subclass (e.g. CropVideo) and confirm parameter order and behaviour are unchanged
- Connect a video to ExtractFrames, provide a frame list (e.g. [1, 10, 50]), run the node, and verify output paths point to valid image files
- Verify ExtractFrames does not show output_frame_rate or processing_speed parameters
 Open the VideoPlayerFrameSelector widget and confirm frame scrubbing works end-to-end

### DIFF
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -60,6 +60,11 @@
       "name": "MaskAdjustmentPreview",
       "path": "widgets/MaskAdjustmentPreview.js",
       "description": "Interactive preview for mask adjustment with timeline scrubbing and dilation/erosion controls."
+    },
+    {
+      "name": "VideoPlayerFrameSelector",
+      "path": "griptape_nodes_library/video/widgets/VideoPlayerFrameSelector.js",
+      "description": "Video player with frame selection capabilities for precise video manipulation nodes."
     }
   ],
   "categories": [
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -768,6 +768,17 @@
       }
     },
     {
+      "class_name": "VideoFrameExtractor",
+      "file_path": "griptape_nodes_library/video/extract_frames.py",
+      "metadata": {
+        "category": "video",
+        "description": "Import video, inspect frame-by-frame, and extract selected frames as images.",
+        "display_name": "Video Frame Extractor",
+        "icon": "film",
+        "group": "edit"
+      }
+    },
+    {
       "class_name": "SeedanceVideoGeneration",
       "file_path": "griptape_nodes_library/video/seedance_video_generation.py",
       "metadata": {
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -775,7 +775,7 @@
       "metadata": {
         "category": "video",
         "description": "Import video, inspect frame-by-frame, and extract selected frames as images.",
-        "display_name": "Video Frame Extractor",
+        "display_name": "Extract Video Frames",
         "icon": "film",
         "group": "edit"
       }
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -768,7 +768,7 @@
       }
     },
     {
-      "class_name": "VideoFrameExtractor",
+      "class_name": "ExtractFrames",
       "file_path": "griptape_nodes_library/video/extract_frames.py",
       "metadata": {
         "category": "video",
--- a/griptape_nodes_library/video/base_video_input_node.py
+++ b/griptape_nodes_library/video/base_video_input_node.py
@@ -109,11 +109,7 @@ class BaseVideoInputNode(SuccessFailureNode, ABC):
 
     @abstractmethod
     def _get_output_file_default_filename(self) -> str:
-        """Return the default filename for the output_file parameter.
-
-        Subclasses may declare OUTPUT_FILE_DEFAULT_FILENAME as a class variable
-        instead of overriding this method.
-        """
+        """Return the default filename for the output_file parameter."""
 
     def _setup_logging_group(self) -> None:
         with ParameterGroup(name="Logs") as logs_group:
--- a/griptape_nodes_library/video/base_video_input_node.py
+++ b/griptape_nodes_library/video/base_video_input_node.py
@@ -1,0 +1,277 @@
+import json
+import subprocess
+import tempfile
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import Any
+
+# static_ffmpeg is dynamically installed by the library loader at runtime
+# into the library's own virtual environment, but not available during type checking
+import static_ffmpeg.run  # type: ignore[import-untyped]
+from griptape.artifacts.video_url_artifact import VideoUrlArtifact
+from griptape_nodes.exe_types.core_types import Parameter, ParameterGroup, ParameterMode
+from griptape_nodes.exe_types.node_types import SuccessFailureNode
+from griptape_nodes.exe_types.param_components.project_file_parameter import ProjectFileParameter
+from griptape_nodes.exe_types.param_types.parameter_video import ParameterVideo
+from griptape_nodes.files.file import File
+
+from griptape_nodes_library.utils.file_utils import generate_filename
+from griptape_nodes_library.utils.video_utils import (
+    detect_video_format,
+    dict_to_video_url_artifact,
+    to_video_artifact,
+    validate_url,
+)
+
+
+class BaseVideoInputNode(SuccessFailureNode, ABC):
+    """Abstract base for nodes that take a video as input.
+
+    Provides video input parameter, ffprobe utilities, URL validation, logging,
+    status parameters, and temp file helpers — without assuming video output.
+    Subclass this when your node produces something other than a video (images,
+    audio, metadata, etc.).  For video-to-video processing, subclass
+    BaseVideoProcessor instead.
+    """
+
+    DEFAULT_FRAME_RATE = 30.0
+    DEFAULT_WIDTH = 1920
+    DEFAULT_HEIGHT = 1080
+    DEFAULT_DURATION = 0.0
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+        self.add_parameter(
+            ParameterVideo(
+                name="video",
+                tooltip="The video to process",
+                clickable_file_browser=True,
+                ui_options={
+                    "expander": True,
+                    "display_name": "Video or Path to Video",
+                },
+                converters=[self._convert_video_input],
+            )
+        )
+
+        self._setup_custom_parameters()
+
+        # Hook: lets BaseVideoProcessor insert frame-rate / speed params here,
+        # between custom params and the primary output param, preserving UI order.
+        self._setup_pre_output_parameters()
+
+        self._register_primary_output_parameter()
+
+        # Separate hook so subclasses can co-locate their output file setup with
+        # _register_primary_output_parameter() if they prefer, while still keeping
+        # the two steps distinct in __init__.
+        self._register_output_file_parameter()
+
+        self._setup_logging_group()
+
+        self._create_status_parameters(
+            result_details_tooltip="Details about the processing operation result",
+            result_details_placeholder="Details on the processing attempt will be presented here.",
+        )
+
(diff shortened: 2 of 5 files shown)


griptape_nodes_library/video/base_video_input_node.py
```diff
@@ -0,0 +1,305 @@
...
+            exceptions.extend(base_exceptions)
+
+        # Add custom validation from subclasses
+        custom_exceptions = self._validate_custom_parameters()
+        if custom_exceptions:
+            exceptions.extend(custom_exceptions)
+
+        return exceptions if exceptions else None
+
+    def _generate_filename(self, suffix: str = "", extension: str = "") -> str:
```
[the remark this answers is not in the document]
alex-rumsey-foundry: That's how it was before me, I can't change the interface without rewriting other nodes. Ideally there should be a verification with throwing an exception for cases with incorrect/missing args.

griptape_nodes_library/video/extract_frames.py
```diff
@@ -0,0 +1,393 @@
...
+        except Exception as e:
+            error_message = str(e)
+            msg = f"{self.name}: Error extracting frames: {error_message}"
+            self.append_value_to_parameter("logs", f"ERROR: {msg}\n")
+            self._set_status_results(was_successful=False, result_details=f"Frame extraction failed: {error_message}")
+            self._handle_failure_exception(ValueError(msg))
+
+    def _run_extraction(self, input_url: str) -> None:
+        saved_paths = self._extract_all_frames(input_url)
+        self.parameter_output_values["output_paths"] = [str(p) for p in saved_paths]
```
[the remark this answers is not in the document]
alex-rumsey-foundry: Was not included in commit, now implemented

griptape_nodes_library/video/widgets/VideoPlayerFrameSelector.js
```diff
@@ -0,0 +1,1257 @@
...
+    overlay.textContent = `Video could not be loaded${codeMsg}.`;
+    overlay.style.display = 'flex'; updateUi();
+  });
+
+  /* ── FPS input ──────────────────────────────────────────────────────────── */
+
+  function commitFps(inputEl, mirrorEl) {
+    const v = Number(inputEl.value);
+    if (Number.isFinite(v) && v >= MIN_FPS && v <= MAX_FPS) { fps = v; mirrorEl.value = String(fps); }
+    else inputEl.value = String(fps);
```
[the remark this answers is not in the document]
alex-rumsey-foundry: nativeFps is not meant to be updated, it's NATIVE, lol

griptape_nodes_library/video/extract_frames.py
```diff
@@ -0,0 +1,393 @@
...
+        """Extract each requested frame and return their saved paths."""
+        self._validate_url_safety(input_url)
+
+        output_format: str = self.get_parameter_value("output_format") or "png"
+        output_prefix: str = self.get_parameter_value("output_prefix") or "frame"
+        frame_padding: int = self.get_parameter_value("frame_padding") or 6
+        output_dir_str: str = self.get_parameter_value("output_dir") or ""
+
+        output_dir = pathlib.Path(output_dir_str) if output_dir_str else self._output_file.build_file().parent
+        output_dir.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
alex-rumsey-foundry: completely updated in latest commit

griptape_nodes_library/video/base_video_input_node.py
```diff
@@ -0,0 +1,305 @@
...
+from griptape_nodes_library.utils.file_utils import generate_filename
+from griptape_nodes_library.utils.video_utils import (
+    detect_video_format,
+    dict_to_video_url_artifact,
+    to_video_artifact,
+    validate_url,
+)
+
+
+class BaseVideoInputNode(SuccessFailureNode, ABC):
```
kateforsberg13, line 27: If this is a repeatable pattern for many nodes, we could even maybe put this Base node in the engine. But that just depends on if you think this would be used for nodes in many libraries, not just this one

griptape_nodes_library/video/base_video_processor.py
```diff
@@ -1,38 +1,26 @@
...
-    dict_to_video_url_artifact,
-    to_video_artifact,
-    validate_url,
-)
+from griptape_nodes_library.video.base_video_input_node import BaseVideoInputNode
 
 
-class BaseVideoProcessor(SuccessFailureNode, ABC):
-    """Base class for video processing nodes with common functionality."""
+class BaseVideoProcessor(BaseVideoInputNode):
```
kateforsberg13, line 14: This looks good so far. Just asking - what is the benefit of now having BaseVideoInputNode and BaseVideoProcessor? What does BaseVideoProcessor provide that BaseVideoInputNode doesn't, and will there be situations where BaseVideoInputNode is extended and it is not a BaseVideoProcessor?

griptape_nodes_library/video/extract_frames.py
```diff
@@ -0,0 +1,425 @@
...
+            try:
+                n = int(token)
+            except ValueError:
+                continue
+            if n >= 1:
+                result.add(n)
+    return sorted(result)
+
+
+class VideoFrameExtractor(BaseVideoInputNode):
```
kateforsberg13: this is kind of a silly one - but maybe rename the file to be video_frame_extractor.py, so it's clear in the standard library for devs who are looking to find the code or make updates.

griptape_nodes_library/video/extract_frames.py
```diff
@@ -0,0 +1,425 @@
...
+from griptape_nodes.traits.file_system_picker import FileSystemPicker
+import griptape_nodes.traits.widget as widget
+
+from griptape_nodes_library.utils.video_utils import seconds_to_ts
+from griptape_nodes_library.video.base_video_input_node import BaseVideoInputNode
+
+logger = logging.getLogger(__name__)
+
+
+def parse_frame_string(frame_str: str) -> list[int]:
```
kateforsberg13, line 35: this is good, but strings can be kind of brittle, especially when they aren't in the format we want. Is there a better way for users to select frames? I liked your demo during CX experience, but was just wondering if you played with anything else.

griptape_nodes_library/video/extract_frames.py
```diff
@@ -0,0 +1,425 @@
...
+        if metadata:
+            node_metadata.update(metadata)
+        super().__init__(name=name, metadata=node_metadata, **kwargs)
+        self.remove_parameter_element_by_name("video")
+        # self.set_initial_node_size(width=980, height=480)
+        self.hide_parameter_by_name("every_n")
+
+    # ── Parameter setup ────────────────────────────────────────────────────────
+
+    def _get_output_file_default_filename(self) -> str:
```
kateforsberg13, line 89: I think we mentioned this in CX , but use existing project structure where you can

griptape_nodes_library/video/extract_frames.py
```diff
@@ -0,0 +1,425 @@
...
+                type="int",
+                default_value=self.DEFAULT_FRAME_PADDING,
+                tooltip="Zero-padding width for the frame number in the output filename.",
+                allowed_modes={core_types.ParameterMode.INPUT, core_types.ParameterMode.PROPERTY},
+            )
+        self.add_node_element(settings_group)
+
+    # ── Validation ─────────────────────────────────────────────────────────────
+
+    def validate_before_node_run(self) -> list[Exception] | None:
```
kateforsberg13, line 225: we defined this in base_video_input_node too. Should it call that validate_before_node_run as well, or does this method handle all of those?

griptape_nodes_library/video/extract_frames.py
```diff
@@ -0,0 +1,425 @@
...
+        return None
+
+    def _resolve_output_dir(self, video_url: str) -> pathlib.Path:
+        """Resolve the project output directory: outputs/{NodeName}/{VideoStem}/.
+
+        Falls back to a temp directory when no project is loaded.
+        """
+        video_stem = pathlib.Path(urlparse(video_url).path).stem or "video"
+        video_stem = re.sub(r"[^\w-]", "_", video_stem)
+        sub_dirs = f"{self.name}/{video_stem}"
```
kateforsberg13: are there any existing methods in projects for this type of thing? genuine question - I do not know!

griptape_nodes_library/video/extract_frames.py
```diff
@@ -0,0 +1,425 @@
...
+                self.hide_parameter_by_name("every_n")
+
+        elif parameter.name in ("output_format", "output_prefix", "frame_padding"):
+            self._sync_output_file_display()
+
+        return super().after_value_set(parameter, value)
+
+    # ── Execution ──────────────────────────────────────────────────────────────
+
+    def process(self) -> node_types.AsyncResult[None]:
```
kateforsberg13: it's probably better practice here to define `aprocess` instead of process. That's already an async abstract function, and then we don't need to use the `yield lambda`.

griptape_nodes_library.json
```diff
@@ -60,6 +60,11 @@
       "name": "MaskAdjustmentPreview",
       "path": "widgets/MaskAdjustmentPreview.js",
       "description": "Interactive preview for mask adjustment with timeline scrubbing and dilation/erosion controls."
```
kateforsberg13, line 64: nice! and we should probably update the library version # for when this is released, esp since it edits `base_video_processor`.

kateforsberg13 (changes requested):

griptape_nodes_library/video/base_video_processor.py
alex-rumsey-foundry: The main problem I had with BaseVideoProcessor is that it assumed the output of the process is always a video. With plans of supporting the sequences it seems important

griptape_nodes_library/video/extract_frames.py
alex-rumsey-foundry: Good point, definitely not silly, I think I renamed the class several times, but never renamed the module. Thanks!

shhlife:
@alex-rumsey-foundry this is very cool!

I had a few quick notes when trying it out:

1. Is it possible to have the Video Frame Extractor widget scale vertically with the node? Right now it scales horizontally.. but the video itself doesn't scale, meaning you can end up with lots of space below it when trying to see all the node parameters as you can see below:

image

2. When I ran the node, I ended up with files named `frame0001.png` instead of `frame.0001.png`, even though the output showed the `.` in the name.

3. After running the node, the engine disconnected and I got the following:
```
[16:53:11] WARNING  Sending large WebSocket message: type=success_result
                    (GetAllNodeInfoResultSuccess), size=5404077 bytes. Large
                    messages can saturate the send buffer and cause connected
                    clients (e.g. the editor) to stall or disconnect.
```

I think we're sending something back to the editor we don't need to? @collindutter or @cjkindel do you have any ideas?

4. The video viewer UI is really cool, and I love seeing the frame indicator! However, I don't know how to click and add markers? Is it possible to show some UI in the little bookmark area and make it so single-clicking in that area will create the bookmarked frame instead of double-clicking? Creating a hover state there would be helpful, too.

5. Some tooltip or over the hover area would be useful, too
6. Is there a way to "select" a bookmarked frame and delete it? I tried & hit delete and it deleted my node. :(
7. When I deleted the node and then re-created it.. it loaded with the video from the previous node already loaded - that seems like an odd bug?
8. I LOVE the previous & next frame buttons - can we also have buttons for prev-next bookmarked frame?
9. Can we have hotkeys for both of the above?

thanks so much for such a great node!

griptape_nodes_library/video/extract_frames.py
alex-rumsey-foundry: Good point! Done.

PR #174, opened 2026-04-29
feat: add Claude Opus 4.7 support to AnthropicPrompt node

ian-massingham:
## Summary

- Adds `claude-opus-4-7` to the `AnthropicPrompt` node model choices
- Hides `temperature`, `top_p`, and `top_k` UI controls when Opus 4.7 is selected, since [Anthropic deprecated all sampling parameters for this model](https://platform.claude.com/docs/en/about-claude/models/migration-guide)
- Strips those params from the driver constructor kwargs when Opus 4.7 is selected

## Dependency

The node-side changes are UX only (hiding inapplicable controls). The fix that actually prevents the 400 error at runtime is in the framework:

> **Requires:** griptape-ai/griptape#2163

This PR should not be merged until that framework PR is merged and the `griptape` dependency is updated.

## Test plan

- [ ] Select `claude-opus-4-7` in the AnthropicPrompt node — temperature, top_p, top_k controls should be hidden
- [ ] Run a workflow with `claude-opus-4-7` (requires framework fix) — should succeed without 400 errors
- [ ] Switch back to `claude-sonnet-4-6` — sampling controls should reappear


griptape_nodes_library.json
```diff
@@ -28,7 +28,7 @@
   "metadata": {
     "author": "Griptape, Inc",
     "description": "Default nodes for Griptape Nodes",
-    "library_version": "0.72.0",
+    "library_version": "0.72.1",
```
collindutter, line 31: Please don't bump library version.

collindutter (changes requested):
Do we need a griptape version bump?

PR #181, opened 2026-04-30
fix(ltx_video_retake): send required resolution parameter

### LINKED ISSUE #179: LTX Video Retake fails on ltx-2-3-pro: missing required 'resolution' parameter
## Problem

The LTX Video Retake node does not send a `resolution` field in its API request. The `ltx-2-3-pro:retake` endpoint rejects requests without it:

```
HTTP 400 - {"error":"invalid input for model ltx-2-3-pro:retake.",
            "provider_response":"resolution parameter is required for retake generation"}
```

## User-visible symptom

`LTXVideoRetake` node errors out when the selected model is `LTX 2.3 Pro`, with the message:

> LTX Video Retake: LTX Video Retake request failed: invalid input for model ltx-2-3-pro:retake.

## Expected behavior

- The node should expose a `resolution` parameter (matching the set supported by the other LTX nodes: `1920x1080`, `2560x1440`, `3840x2160`).
- Ideally it is auto-populated from the input video's actual dimensions, since a retake should preserve the source resolution.
- The resolution should be included in the request payload.

## Context

The sibling LTX nodes (`LTXTextToVideoGeneration`, `LTXImageToVideoGeneration`, `LTXAudioToVideoGeneration`) already expose and send a `resolution` parameter. `LTXVideoRetake` is the only one that doesn't.

collindutter:
## Summary

Add the missing `resolution` parameter to `LTXVideoRetake` so requests against `ltx-2-3-pro:retake` no longer fail with HTTP 400.

## Symptom

```
HTTP error: 400 - {"error":"invalid input for model ltx-2-3-pro:retake.",
                   "provider_response":"resolution parameter is required for retake generation"}
```

## Approach

- **New `resolution` `ParameterString`** with the same choices as the other LTX nodes (`1920x1080`, `2560x1440`, `3840x2160`), defaulting to `1920x1080`. User-visible, user-overridable.
- **Auto-detect from input video.** The node already calls `ffprobe` to read the input video's duration for the retake-segment range. That helper is split into a shared `_get_video_stream_info()` that returns duration + width/height in one call. When a video is attached, the resolution parameter is auto-snapped to the closest supported size by pixel count (`_snap_to_supported_resolution`).
- **Payload** now includes `"resolution": ...`, validated against the supported set before sending.

## Files touched

- `griptape_nodes_library/video/ltx_video_retake.py`

## Verification

- File parses and passes `ruff check`.
- Behavior on an attached video: 1080p → `1920x1080`, 1440p → `2560x1440`, 4K → `3840x2160`; users can override via the UI.

Closes #179

### DIFF
--- a/griptape_nodes_library/video/ltx_video_retake.py
+++ b/griptape_nodes_library/video/ltx_video_retake.py
@@ -33,6 +33,9 @@ MODEL_MAPPING = {
     "LTX 2.3 Pro": "ltx-2-3-pro",
 }
 
+SUPPORTED_RESOLUTIONS = ("1920x1080", "2560x1440", "3840x2160")
+DEFAULT_RESOLUTION = "1920x1080"
+
 
 class LTXVideoRetake(GriptapeProxyNode):
     """Regenerate a segment of an existing video using LTX AI via Griptape Cloud model proxy.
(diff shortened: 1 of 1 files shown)


shhlife:
@collindutter and @andrewfrench this worked for me.. can we get it in?

cjkindel (approved):

PR #191, opened 2026-05-06
fix: save_image conversion to inferred file format from output file name (#190)

alex-rumsey-foundry:
## Fix `SaveImage` writing raw bytes without format conversion

Fixes #190 

### Problem

`SaveImage` was occasionally saving JPEG payloads with a `.png` extension (and vice-versa), producing files that image viewers and downstream nodes would misidentify or reject.

The root cause: `process()` called `image_artifact.to_bytes()` — which returns the raw internal bytes with no format conversion — and wrote those bytes directly to disk, completely ignoring the extension of the configured output file. The `_extract_format_from_artifact` helper existed to detect the incoming format but was dead code; it was never called during execution.

### Fix

- Added `_get_target_pil_format()` which resolves the intended output format from the `output_file` parameter (e.g. `griptape_nodes.png` → `PNG`, `output.jpg` → `JPEG`).
- Before writing, the target format is validated against PIL's supported formats via `validate_pil_format()`. If the extension is unsupported (e.g. `.tga`), the node now fails with a descriptive error instead of silently writing garbage bytes.
- Image bytes are opened with PIL and re-encoded to the target format using the existing `image_to_bytes()` utility. RGBA/LA/P images are converted to RGB before JPEG encoding, since JPEG has no alpha channel support.
- `_extract_format_from_artifact` is now wired in to log the detected source format at debug level, making format-mismatch issues easier to diagnose.

### DIFF
--- a/griptape_nodes_library/image/save_image.py
+++ b/griptape_nodes_library/image/save_image.py
@@ -1,5 +1,6 @@
 from enum import StrEnum, auto
 from io import BytesIO
+from pathlib import Path
 from typing import Any
 
 from griptape.artifacts import ImageArtifact, ImageUrlArtifact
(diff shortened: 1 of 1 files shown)


kateforsberg13 (approved):
looks good to me!

PR #208, opened 2026-05-11
feat: add ConvertSpzToPly node for SPZ → PLY Gaussian splat conversion

kateforsberg13:
## Summary

Adds a `ConvertSpzToPly` node under the **Splat** category (group: Transform). Wire any SPZ splat in (e.g. from `WorldLabsWorldGeneration`) and get a standard 3DGS-compatible PLY back out
## Why

Marble's `assets.splats` response only contains `spz_urls` — there is no server-side PLY. Users who want to work with their generated worlds in downstream DCC tools (Blender, Unreal, etc.) currently have to round-trip through the third-party `spz-to-ply.netlify.app` site that Marble's docs link to. This node brings that conversion in-process.

A separate node (rather than a `splat_format` knob on `WorldLabsWorldGeneration`) keeps responsibilities clean: the generator stays focused on the Marble API, the converter is reusable with any SPZ source, and users who only want SPZ never pay the conversion cost.

## Implementation notes
- **Pure-Python decoder.** Niantic's `spz` bindings require git+cmake at install time and have no cross-platform wheels; they also don't expose a `save_to_ply` function. A pure-Python decoder (~200 LOC of numpy) avoids both problems. Only stdlib + numpy + `zstandard` (already a transitive dep) are needed.
- **Dual-format support.** Handles both legacy SPZ (gzip-wrapped, versions 1–3, the format Marble currently returns) and NGSP v4 (plaintext header + ZSTD per-stream). The parser auto-detects by magic bytes.
- **Pass-through for existing PLY.** If the upstream artifact's `meta["format"] == "ply"`, the node short-circuits and re-emits without re-conversion.
- `WorldLabsWorldGeneration` now stamps `meta["format"] = "spz"` on its splat outputs so downstream nodes (including this one) don't have to guess.


cjkindel (approved):

griptape_nodes_library/splat/convert_spz_to_ply.py
```diff
@@ -0,0 +1,399 @@
...
+        except Exception as exc:
+            raise RuntimeError(f"Failed to save PLY: {exc}") from exc
+
+        upstream_meta["format"] = "ply"
+        self.parameter_output_values["splat_out"] = SplatUrlArtifact(
+            value=saved.location,
+            meta=upstream_meta,
+        )
+
+    @staticmethod
```
alex-rumsey-foundry: Does it have to be a static method rather than a function?

griptape_nodes_library/splat/convert_spz_to_ply.py
```diff
@@ -0,0 +1,399 @@
...
+    def _extract_splat_url(raw: Any) -> str | None:
```
alex-rumsey-foundry: More below, if we expect certain input types, we don't need to do that. Also could use some user feedback if we can't find a URL. AFAICS the current implementation (lines 358-359) just quietly finishes the node execution.

griptape_nodes_library/splat/convert_spz_to_ply.py
```diff
@@ -0,0 +1,399 @@
...
+        super().__init__(name, metadata)
+        self.category = "Splat"
+        self.description = "Convert an SPZ Gaussian splat to PLY for broader software compatibility."
+
+        self.add_parameter(
+            ParameterSplat(
+                name="splat_in",
+                tooltip="Wire an SPZ splat (e.g. from WorldLabsWorldGeneration) or pick a local .spz file.",
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+                accept_any=True,
```
alex-rumsey-foundry: Is it a common practice to allow "Any" as input for a node? Shouldn't we limit this to `str|ThreeDUrlArtifact` (I'm not sure if we have one)?

alex-rumsey-foundry (commented):
I'm not keen on PLY format spec, so can't properly verify if the implementation is correct. I guess we'll have to trust the coding agent here :(

My only comments is about the node input/behaviour. Not a request of change given the critical status of the ticket, but maybe worth taking a look to maintain a better quality of the flow.

griptape_nodes_library/splat/convert_spz_to_ply.py
kateforsberg13: Updated to remove static method

griptape_nodes_library/splat/convert_spz_to_ply.py
kateforsberg13: Updated to add that. Sometimes connections can be finicky, so we'll add "any" but you're right, this is a good example. Model oversight

PR #238, opened 2026-05-14
refactor(media): consolidate duplicated media input/output helpers

collindutter:
Closes #237.

PR #236 had to apply the same one-line dict-input fix to three byte-identical copies of `_coerce_video_url_or_data_uri`, and a parallel `output.video_url` fix to two siblings whose helpers happened to be named differently. Both bugs would have been fixed once under a shared helper, and any future provider node added to this library compounds the duplication further.

The new `griptape_nodes_library.media` module exposes `coerce_media_url_or_data_uri(val, *, kind)` and `prepare_media_data_uri(val, *, kind, ...)`, parameterized on the MIME prefix and default extension for `image` / `video` / `audio`. The serialized-dict branch that was unique to `Seedance20VideoGeneration._coerce_image_url_or_data_uri` is folded in so every caller benefits, including for `AudioArtifact` dicts. `prepare_media_data_uri` accepts a `fallback_mime` override because some provider APIs prefer `image/jpeg` over the canonical `image/png` when the loader cannot otherwise determine a type.

The input-extractor overloads in `ltx_video_retake` and `ltx_video_to_video_hdr` are renamed to `_extract_input_video_url` so they no longer share a name with the JSON response extractors elsewhere (`_extract_video_url` / `_extract_result_video_url`). The response extractors themselves are intentionally left per-node, since each provider's response shape really is different.

Per-node `_prepare_*_async` wrappers that genuinely diverge from the standard shape (Kling stripping the data-URI prefix, Kling Omni returning raw base64 from `aread_bytes`, LTX audio normalizing the MIME afterward, Seedance 2.0 short-circuiting on `asset://`) keep their custom bodies but route through the shared coercer. Tests for the helpers move to `tests/unit/media/test_coercion.py` and exercise the shared functions directly, the dict path for both image and audio, and the renamed LTX input extractors.

Refs: #236, https://github.com/orgs/griptape-ai/discussions/2178


griptape_nodes_library/media/coercion.py
```diff
@@ -0,0 +1,172 @@
...
+from griptape_nodes.files.file import File, FileLoadError
+
+logger = logging.getLogger("griptape_nodes")
+
+MediaKind = Literal["image", "video", "audio"]
+
+_DEFAULT_EXTENSION: dict[MediaKind, str] = {
+    "image": "png",
+    "video": "mp4",
+    "audio": "mpeg",
```
cjkindel, line 38: really mpeg over mp3?

collindutter: https://stackoverflow.com/questions/10688588/which-mime-type-should-i-use-for-mp3

Also all nodes that this logic was extracted from use mpeg.

cjkindel: I was asking on extension, not mime type

collindutter: Ah, this is a poorly named variable. It's only used for mime type purposes. Updated.

cjkindel (approved):

PR #243, opened 2026-05-15
fix(wan-animate): upload inputs as public URLs and read DashScope's nested results shape

collindutter:
WAN Animate has been failing for two unrelated reasons since the Feb 4 sync from `griptape-nodes@23e7ef1` (`528f139`, first shipped in v0.63.9):

1. `_build_payload` switched from `get_public_url_for_parameter()` to `prepare_media_data_uri`, so `input.image_url` and `input.video_url` started going out as `data:` URIs. The Animate endpoint (`services/aigc/image2video/video-synthesis`) is one of the few DashScope endpoints the cloud proxy hits via raw `requests.post` rather than the DashScope SDK, so the field is forwarded verbatim and the provider rejects the data URI with `code InvalidVideo.FileFormat: Invalid video type. Only mp4/mov/avi is supported.` Restoring the `PublicArtifactUrlParameter` upload path matches what `Seedance20VideoGeneration`, `KlingMotionControl`, and `SeedVRVideoUpscale` already do for the same reason; cleanup of the short-lived asset still runs in the existing `finally` block via `delete_uploaded_artifact()`.

2. `_extract_status` and `_extract_video_url` were probing flat shapes that don't match what DashScope actually returns. The documented response is `output.task_status` and `output.results.video_url` (see <https://help.aliyun.com/zh/model-studio/wan2-2-animate-mix-api>); without those lookups, a successful generation surfaced as "Generation completed but no video URL was found" and a FAILED task fell through to the same misleading branch instead of producing the provider error. The legacy `output.video_url`, top-level `results.video_url`, and top-level `video_url` shapes are kept as fallbacks so older / sibling response variants still resolve.

image

### DIFF
--- a/griptape_nodes_library/video/wan_animate_generation.py
+++ b/griptape_nodes_library/video/wan_animate_generation.py
@@ -18,7 +18,6 @@ from griptape_nodes.exe_types.param_types.parameter_video import ParameterVideo
 from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 from griptape_nodes.traits.options import Options
 
-from griptape_nodes_library.media import prepare_media_data_uri
 from griptape_nodes_library.proxy import GriptapeProxyNode
 from griptape_nodes_library.utils.video_utils import get_video_duration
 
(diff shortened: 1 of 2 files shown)


cjkindel (approved):
Turns out Claude may not know everything initially

PR #252, opened 2026-05-18
feat(proxy): expose generation timeout as user-configurable parameter

kateforsberg13:
## Summary

closes #235 
- Adds a `timeout` parameter (seconds) to every proxy node via the `GriptapeProxyNode` base class — no per-subclass changes needed
- Default is seeded from each class's `DEFAULT_MAX_ATTEMPTS * DEFAULT_POLL_INTERVAL`, preserving existing behaviour (600s for most nodes, 1200s for LTX/Minimax)
- Setting `timeout=0` disables the cap entirely — the node will poll until the job completes or fails
- Removes the now-redundant local `timeout` overrides from `SeedVRImageUpscale` and `SeedVRVideoUpscale`
- Polling loop refactored to avoid mutating class attributes on the instance

## Test plan

- [x] All 48 new unit tests in `tests/unit/test_griptape_proxy_node_polling.py` pass
  - Parametrized check that every proxy node has a `timeout` param with the correct default
  - `timeout=10` stops polling after 2 attempts (10s / 5s interval) with a clear timeout message
  - `timeout=0` polls until completion without firing the timeout branch
  - Class attribute `DEFAULT_MAX_ATTEMPTS` is never mutated after a run
- [x] Open a proxy node in the UI — `timeout` field visible as a top-level parameter, default 600
- [x] `LTXTextToVideoGeneration` shows default 1200
- [x] `SeedVRImageUpscale` shows default 600 with no duplicate field
- [x] Set `timeout=30` on a slow model and confirm it times out with `Generation timed out after 30 seconds waiting for result.`

Closes #235

### DIFF
--- a/griptape_nodes_library/image/seedvr_image_upscale.py
+++ b/griptape_nodes_library/image/seedvr_image_upscale.py
@@ -16,7 +16,6 @@ from griptape_nodes.exe_types.param_components.seed_parameter import SeedParamet
 from griptape_nodes.exe_types.param_types.parameter_dict import ParameterDict
 from griptape_nodes.exe_types.param_types.parameter_float import ParameterFloat
 from griptape_nodes.exe_types.param_types.parameter_image import ParameterImage
-from griptape_nodes.exe_types.param_types.parameter_int import ParameterInt
 from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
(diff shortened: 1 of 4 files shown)


collindutter (approved):
Nice!

PR #257, opened 2026-05-19
Make Adjust Mask Size dramatically faster

shhlife:
## Summary

- Replaced the Python frame-extract → PIL/numpy process → reassemble pipeline in `AdjustMaskSize` with a single FFmpeg invocation using the built-in `dilation`/`erosion` video filters
- Removed `json`, `numpy`, and `PIL` imports — no longer needed
- Deleted five methods that are now dead code: `_get_video_properties`, `_extract_frames`, `_adjust_frames`, `_dilate_binary`, `_erode_binary`, `_reassemble_video`
- File reduced from ~520 lines to ~298 lines

## How it works

For a given adjustment value N, the new code builds a chained FFmpeg filter string:

```
format=gray,dilation,dilation,...   # N times for expansion
format=gray,erosion,erosion,...     # N times for shrinkage
```

FFmpeg applies all filters in a single streaming decode→filter→encode pass, running entirely in C with no per-frame disk I/O or Python overhead.

## Performance

Previously: extract every frame as a PNG → read each PNG → numpy → PIL MaxFilter/MinFilter → write PNG → re-encode video (N+2 disk round-trips per frame).

Now: one FFmpeg subprocess call.

## Trade-off

FFmpeg's `dilation`/`erosion` filters use a 3×3 neighborhood per step, so chaining N times produces a diamond-shaped structuring element rather than the circular kernel used previously. The difference is imperceptible for typical mask adjustment values (1–10px).

## Test plan

- [x] Run `AdjustMaskSize` node with a mask video and positive adjustment — output mask should be dilated
- [x] Run with negative adjustment — output mask should be eroded
- [x] Verify adjustment = 0 passes the input through unchanged (short-circuits before FFmpeg)
- [x] Verify cleanup: no temp files left behind after processing

### DIFF
--- a/griptape_nodes_library/video/adjust_mask_size.py
+++ b/griptape_nodes_library/video/adjust_mask_size.py
@@ -245,13 +245,20 @@ class AdjustMaskSize(DataNode):
 
             cmd = [
                 ffmpeg_path,
-                "-i", video_url,
-                "-vf", vf,
-                "-c:v", "libx264",
-                "-pix_fmt", "yuv420p",
-                "-preset", "medium",
-                "-crf", "18",
-                "-y", str(output_video),
+                "-i",
+                video_url,
+                "-vf",
+                vf,
+                "-c:v",
+                "libx264",
+                "-pix_fmt",
+                "yuv420p",
+                "-preset",
+                "medium",
+                "-crf",
+                "18",
+                "-y",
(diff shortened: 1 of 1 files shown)


collindutter (approved):
I'll never say no to a PR that deletes more than it adds

PR #270, opened 2026-05-27
Enable retrieval of a generation after client timeout

zachgiordano:
Closes #192

## Summary

Adds a recovery affordance for `GriptapeProxyNode` when a long-running Griptape Cloud generation outlives the client-side polling window. Today a timeout calls `_set_safe_defaults()` and the cloud-side artifact is unrecoverable from the node — the user has to re-run the workflow. After this change, `generation_id` and `generation_status` are preserved across timeout/failure paths and a Refresh button on the node fetches the result on demand.

Screenshot 2026-05-27 at 11 46 05 AM

## Changes

**Base class (`griptape_nodes_library/proxy/griptape_proxy_node.py`)**
- Adds three elements to the existing collapsible **Status** group via an override of `_create_status_parameters`, so subclasses pick them up automatically:
  - `generation_id` (output, hidden by default — surfaces as a port for connections without taking node real estate).
  - `generation_status` (output) — values: `QUEUED`, `RUNNING`, `COMPLETED`, `FAILED`, `ERRORED`, `CANCELLED`, `TIMED_OUT`. Updated on every poll iteration so it's fresh mid-flight.
  - `generation_refresh` (`ParameterButton`) — re-checks status and pulls the result onto the node if completed.
- Polling timeout now preserves `generation_id`, sets `generation_status="TIMED_OUT"`, and writes a `result_details` message pointing the user at the Refresh button (instead of `_set_safe_defaults` blowing away the ID).
- `_handle_terminal_status` for `FAILED`/`ERRORED`/`CANCELLED` similarly preserves `generation_id` and writes the terminal status.
- Refresh flow: `_on_refresh_clicked` runs the async refresh on a worker thread (avoids `RuntimeError: This event loop is already running` when the click handler fires inside an existing loop). Single `GET /generations/{id}`; never re-enters the polling loop. Branches on status and only fetches/parses the `/result` endpoint when status is `COMPLETED`.
- `_submit_and_poll` writes `parameter_output_values["generation_id"]` unconditionally (it was previously gated on the key already existing in the dict, which meant the ID wasn't surfaced on first runs).

**Subclasses (44 files)**
Removed the per-subclass `generation_id` `ParameterString` declaration since it now lives on the base. No other subclass code changes.

**Tests (`tests/unit/test_griptape_proxy_node_refresh.py`)** — 6 new tests:
- Status group structure (`generation_id`, `generation_status`, `generation_refresh` all present).
- Polling timeout preserves `generation_id` and writes `TIMED_OUT`.
- `_handle_terminal_status` preserves `generation_id` on `FAILED`.
- `_refresh_async` with no `generation_id` reports gracefully without making HTTP calls.
- `_refresh_async` while `RUNNING` updates status and does not invoke `_parse_result`.
- `_refresh_async` on `COMPLETED` hits both the status and `/result` endpoints and forwards the payload to `_parse_result`.

## Test plan

- [x] `uv run ruff check` clean
- [x] `uv run ruff format` clean
- [x] `uv run pytest tests/unit/test_griptape_proxy_node_*.py` — 138 passed (was 132 before this PR's tests), 9 pre-existing failures unchanged
- [x] Manual: run a proxy node with `timeout=2` so it times out before completion. Confirm node fails with timeout message, `generation_id` is populated on the output port, `generation_status="TIMED_OUT"`, and the Refresh button is visible inside the Status group.
- [x] Manual: wait for the cloud generation to finish, click Refresh. Confirm `generation_status` flips to `COMPLETED`, the result outputs populate, and `result_details` reflects success.
- [x] Manual: click Refresh while still `RUNNING`. Confirm status updates and no result is fetched.
- [ ] Manual: trigger a failure (invalid model id). Confirm `generation_id` is preserved and error details are surfaced via `result_details`.

### DIFF
--- a/griptape_nodes_library/audio/eleven_labs_music_generation.py
+++ b/griptape_nodes_library/audio/eleven_labs_music_generation.py
@@ -118,16 +118,6 @@ class ElevenLabsMusicGeneration(GriptapeProxyNode):
 
         # OUTPUTS
         self.add_parameter(
-            ParameterString(
-                name="generation_id",
-                tooltip="Generation ID from the API",
-                allowed_modes={ParameterMode.OUTPUT},
-                hide_property=True,
-                hide=True,
-            )
-        )
-
-        self.add_parameter(
             ParameterAudio(
                 name="audio_url",
                 tooltip="Generated music audio as URL artifact",
--- a/griptape_nodes_library/audio/eleven_labs_sound_effect.py
+++ b/griptape_nodes_library/audio/eleven_labs_sound_effect.py
@@ -88,16 +88,6 @@ class ElevenLabsSoundEffectGeneration(GriptapeProxyNode):
 
         # OUTPUTS
         self.add_parameter(
-            ParameterString(
-                name="generation_id",
-                tooltip="Generation ID from the API",
-                allowed_modes={ParameterMode.OUTPUT},
-                hide_property=True,
-                hide=True,
-            )
-        )
-
-        self.add_parameter(
             ParameterAudio(
                 name="audio_url",
                 tooltip="Generated sound effect audio as URL artifact",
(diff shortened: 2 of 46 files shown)


shhlife (approved):
this worked great for me!

PR #273, opened 2026-05-27
feat: add LoadSplat node and SplatViewer widget

kateforsberg13:
## Summary
Screenshot 2026-05-27 at 4 32 44 PM

- Adds `LoadSplat` node: aggregates wired splats from upstream (e.g. `WorldLabsWorldGeneration`'s three resolution outputs) and exposes a priority-based pass-through output (`full_res > 500k > 100k`). This matches the World Labs Generation pattern (100k, 500k, full res). 
- Adds `SplatViewer.js` widget: inline Three.js + Spark.js 3D Gaussian splat renderer with orbit/pan/zoom controls, resolution dropdown, and LoD knobs
- Registers both in `griptape_nodes_library.json` and `splat/__init__.py`

## Test plan

- [ ] Drop a `LoadSplat` node on the canvas — confirm it appears under category `Splat` / group `Input/Output`
- [ ] Wire `WorldLabsWorldGeneration` outputs to `splat_100k`, `splat_500k`, `splat_full_res` inputs — confirm `viewer_state` populates and `SplatViewer` widget renders the splat inline
- [ ] Confirm `splat_out` passes through the highest-resolution wired input
- [ ] `uv run pytest tests/unit/splat/` — all 55 tests pass

### DIFF
--- a/griptape_nodes_library/splat/load_splat.py
+++ b/griptape_nodes_library/splat/load_splat.py
@@ -1,0 +1,244 @@
+"""LoadSplat — wired splat aggregator with self-contained inline viewer."""
+
+import json
+import logging
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import BaseNode, DataNode
+from griptape_nodes.retained_mode.events.parameter_events import SetParameterValueRequest
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+from griptape_nodes.traits.widget import Widget
+
+from griptape_nodes_library.splat.parameter_splat import ParameterSplat
+from griptape_nodes_library.splat.splat_artifact import SplatUrlArtifact
+
+logger = logging.getLogger("griptape_nodes")
+
+VIEWER_STATE_DEFAULTS: dict[str, Any] = {
+    "splats": {"100k": None, "500k": None, "full_res": None},
+    "defaults": {
+        "flip_coordinates": True,
+        "enable_lod": True,
+        "lod_scale": 1.0,
+        "max_sh": 3,
+    },
+}
+
+INPUT_TO_RESOLUTION_KEY = {
+    "splat_100k": "100k",
+    "splat_500k": "500k",
+    "splat_full_res": "full_res",
+}
+OUTPUT_PRIORITY = ["splat_full_res", "splat_500k", "splat_100k"]
+
+
+class LoadSplat(DataNode):
+    """View Gaussian splats inline.
+
+    Wire one or more SplatUrlArtifacts (typically from WorldLabsWorldGeneration's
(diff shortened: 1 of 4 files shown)


griptape_nodes_library/splat/load_splat.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+        self.category = "Splat"
+        self.description = "Aggregate wired splats and view inline (resolution + LoD selection live in the widget)."
+
+        # --- Wired inputs ---------------------------------------------------
+        self.add_parameter(
+            ParameterSplat(
```
shhlife, line 46: I think this is a UI note - but in the ParameterSplat component, when loading a splat the spinner looks pretty janky.. it's stretched out.. you can see the attached video.

https://github.com/user-attachments/assets/7a5eaab2-7a1a-4bba-99cf-bc30e98093f8

We should log a ticket to fix the display to look like a standard spinner.

griptape_nodes_library/splat/load_splat.py
```diff
@@ -0,0 +1,244 @@
...
+                default_value=json.dumps(VIEWER_STATE_DEFAULTS),
+                tooltip="Aggregated splat URLs + viewer defaults (JSON). Driven by the wired inputs.",
+                allowed_modes={ParameterMode.PROPERTY},
+                traits={Widget(name="SplatViewer", library="Griptape Nodes Library")},
+                ui_options={"display_name": "Viewer", "is_full_width": True},
+            )
+        )
+
+        # --- Pass-through output --------------------------------------------
+        self.add_parameter(
```
shhlife, line 94: This is still showing the clickable file browser - which it shouldn't? again, is this a UI bug?

Image

shhlife (changes requested):
I wonder if the labels and dropdowns could be inline here to reduce vertical screen space?

Image

Also, my viewer in the widget isn't displaying the splat the same as the viewer does.. I need to reverse y/z, and then it's still showing upside down.

Image

Image

griptape_nodes_library/splat/load_splat.py
kateforsberg13: I don't think I can see the video! do you mean in the full screen viewer, or in the node itself?

griptape_nodes_library/splat/load_splat.py
kateforsberg13: Here you should be able to click on that and select a file to upload. So, it should be a clickable file browser. Do you mean it's displaying the wrong UI button?

shhlife (changes requested):
This is so much better, thanks @kateforsberg13 !

Only one small nit - the order of the parameter inputs doesn't match the dropdown order - can we sync those?

Image

cheers! :)

shhlife (approved):
great work!

PR #287, opened 2026-05-29
feat: add frame range splitting to Split Video, add Trim Video node

shhlife:
## Summary

Closes #285, Closes #284

- **Split Video**: Added `split_by` dropdown (`timecode` / `frame range`) so users can split by frame numbers (e.g. `0-100`) instead of requiring SMPTE timecodes. The active input shows/hides automatically.
- **Trim Video**: New node with a single video output, `trim_by` dropdown (`timecode` / `frame range`), and separate `start`/`end` parameters for each mode.
- **DRY refactor**: Extracted shared ffmpeg helpers into `video_utils.py` — `get_ffmpeg_paths()`, `detect_video_properties()`, and `build_video_segment_cmd()` — removing ~100 lines of duplicated logic from `split_video.py`.
- **Bug fix**: Stream copy now uses fast seek (pre-input `-ss`/`-to`) instead of accurate seek. Post-input seek with `-c copy` silently drops all video frames for H.264/MOV files (output was 2 KB container with `video:0KiB`). Also fixed `-map 0` including unsupported `tmcd` timecode data streams that caused MP4 header write failures.

## Examples

image

image

## Test plan

- [x] Split Video — timecode mode: split a video using SMPTE timecodes, verify segments have content
- [x] Split Video — frame range mode: enter 
```
0-100
100-200
```
- [x] Trim Video — timecode mode: trim with start/end timecodes, verify single output with correct content
- [x] Trim Video — frame range mode: trim with start/end frame numbers, verify single output with correct duration
- [x] Switching `split_by` / `trim_by` hides/shows the correct input fields
- [x] Videos with embedded timecode tracks (Apple ProRes, etc.) no longer fail at the MP4 header write step


griptape_nodes_library/video/trim_video.py
```diff
@@ -0,0 +1,258 @@
...
+            out_path = Path(temp_dir) / "trimmed.mp4"
+            cmd = build_video_segment_cmd(ffmpeg_path, input_url, start_sec, end_sec, str(out_path))
+            self.append_value_to_parameter("logs", f"Running ffmpeg command: {' '.join(cmd)}\n")
+
+            try:
+                result = subprocess.run(cmd, capture_output=True, text=True, check=True, timeout=300)  # noqa: S603
+                if result.stderr:
+                    self.append_value_to_parameter("logs", f"FFmpeg stderr: {result.stderr}\n")
+            except subprocess.TimeoutExpired as e:
+                raise ValueError("FFmpeg timed out after 5 minutes") from e
```
cjkindel: SplitVideo puts the nodename in log, may be valuable here (which begs another question, should we have a standardized "run ffmpeg w/ these params" function

griptape_nodes_library/utils/video_utils.py
```diff
@@ -41,6 +41,138 @@
...
+        duration = 0.0
+        if "format" in data and "duration" in data["format"]:
+            duration = float(data["format"]["duration"])
+        elif "duration" in stream:
+            duration = float(stream["duration"])
+
+        return frame_rate, drop_frame, duration
+
+    except Exception:
+        return 24.0, False, 0.0
```
cjkindel: I don't like auto-defaulting to hidden params the user didn't input. My preference is for a node to fail when given illegal params. (not blocking, just opinion)

I really don't like silently doing this. We need at **least** a log here telling the customer their inputs were illegal/ignored

griptape_nodes_library/utils/video_utils.py
```diff
@@ -41,6 +41,138 @@
...
+
+    Stream copy requires fast seek — post-input accurate seek with -c copy silently
+    drops video frames for many codecs/containers (e.g. H.264 in MOV/MP4).
+    Data streams like timecode tracks (tmcd) are excluded since MP4 doesn't support them.
+    """
+    ss = seconds_to_ts(start_sec)
+    to = seconds_to_ts(end_sec)
+    duration = end_sec - start_sec
+
+    if duration >= MIN_SEGMENT_DURATION_FOR_STREAM_COPY and start_sec == 0.0:
```
cjkindel: nit: `start_sec < 1e-5` or something like that is a little safer

griptape_nodes_library/video/trim_video.py
```diff
@@ -0,0 +1,258 @@
...
+                raise ValueError(f"Invalid or unsafe URL: {input_url}")  # noqa: TRY301
+
+            self.append_value_to_parameter("logs", "Detecting video properties...\n")
+            _, ffprobe_path = get_ffmpeg_paths()
+            frame_rate, drop_frame, video_duration = detect_video_properties(input_url, ffprobe_path)
+            self.append_value_to_parameter("logs", f"Frame rate: {frame_rate} fps, Duration: {video_duration:.2f}s\n")
+
+            if trim_by == "frame range":
+                start_frame = self.get_parameter_value("start_frame") or 0
+                end_frame = self.get_parameter_value("end_frame") or 100
```
cjkindel: nit: not a risk here but we have to be careful about evaluating ints like this. `0` would get treated as false and resolve as 100

cjkindel (changes requested):

griptape_nodes_library/utils/video_utils.py
```diff
@@ -41,6 +42,186 @@
...
+VIDEO_DURATION_BUFFER = 0.1  # seconds — trim end slightly inside duration to avoid keyframe issues
+
+
+def get_ffmpeg_paths() -> tuple[str, str]:
+    """Return (ffmpeg_path, ffprobe_path) from static_ffmpeg."""
+    try:
+        return static_ffmpeg.run.get_or_fetch_platform_executables_else_raise()
+    except Exception as e:
+        error_msg = f"FFmpeg not found. Please ensure static-ffmpeg is properly installed. Error: {e!s}"
+        raise ValueError(error_msg) from e
```
SavagePencil: We have these (or comparable ones) in utils/ffmpeg_utils.py

griptape_nodes_library/utils/video_utils.py
```diff
@@ -41,6 +42,186 @@
...
+
+
+def detect_video_properties(
+    input_url: str,
+    ffprobe_path: str,
+    *,
+    log: Callable[[str], None] | None = None,
+) -> tuple[float, bool, float]:
```
SavagePencil: nit: NamedTuples tell you what the tuple fields are for

griptape_nodes_library/utils/video_utils.py
```diff
@@ -41,6 +42,186 @@
...
+        ]
+        result = subprocess.run(cmd, capture_output=True, text=True, check=True, timeout=30)  # noqa: S603
+        data = json.loads(result.stdout)
+
+        if not data.get("streams"):
+            if log:
+                log(
+                    f"WARNING: ffprobe found no video streams — using defaults: {_DEFAULT_MSG}. Timecode/frame-range calculations may be inaccurate.\n"
+                )
+            return _DEFAULTS
```
SavagePencil: is this not a deal killer, if we can't find any streams?

griptape_nodes_library/utils/video_utils.py
```diff
@@ -41,6 +42,186 @@
...
+
+    Stream copy requires fast seek — post-input accurate seek with -c copy silently
+    drops video frames for many codecs/containers (e.g. H.264 in MOV/MP4).
+    Data streams like timecode tracks (tmcd) are excluded since MP4 doesn't support them.
+    """
+    ss = seconds_to_ts(start_sec)
+    to = seconds_to_ts(end_sec)
+    duration = end_sec - start_sec
+
+    if duration >= MIN_SEGMENT_DURATION_FOR_STREAM_COPY and start_sec < 1e-5:
```
SavagePencil: I think Python offers an equivalent to `float.MIN_VAL` so you don't need to define a custom threshold, but maybe this is actually the correct value?

griptape_nodes_library/utils/video_utils.py
```diff
@@ -41,6 +42,186 @@
...
+            "-map",
+            "0:v",
+            "-map",
+            "0:a?",
+            "-c",
+            "copy",
+            "-movflags",
+            "+faststart",
+            output_path,
+        ]
```
SavagePencil: since this looks like a subset of the output from the success path, maybe we start with this and add in just the ones we want for the success path, so we're only barfing out this list of strings in one place?

griptape_nodes_library/video/split_video.py
```diff
@@ -14,17 +11,20 @@
...
 from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
-
-# static_ffmpeg is dynamically installed by the library loader at runtime
-# into the library's own virtual environment, but not available during type checking
 from griptape_nodes.exe_types.param_types.parameter_video import ParameterVideo
 from griptape_nodes.files.file import File
 from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes, logger
+from griptape_nodes.traits.options import Options
 
 from griptape_nodes_library.utils.video_utils import (
```
SavagePencil, line 26: I think we should look at what's in the utils.ffmpeg_utils.py to see what we can consolidate/reduce

griptape_nodes_library/video/split_video.py
```diff
@@ -169,6 +133,16 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
...
         self.add_node_element(logs_group)
 
+    def after_value_set(self, parameter: Parameter, value: Any) -> None:
+        if parameter.name == "split_by":
+            if value == "timecode":
+                self.show_parameter_by_name("timecodes")
+                self.hide_parameter_by_name("frame_ranges")
+            elif value == "frame range":
+                self.hide_parameter_by_name("timecodes")
+                self.show_parameter_by_name("frame_ranges")
```
SavagePencil: paranoia: do a match/case on these strings so you can catch a failure

griptape_nodes_library/video/trim_video.py
```diff
@@ -0,0 +1,253 @@
...
+            if value == "timecode":
+                self.show_parameter_by_name("start_timecode")
+                self.show_parameter_by_name("end_timecode")
+                self.hide_parameter_by_name("start_frame")
+                self.hide_parameter_by_name("end_frame")
+            elif value == "frame range":
+                self.hide_parameter_by_name("start_timecode")
+                self.hide_parameter_by_name("end_timecode")
+                self.show_parameter_by_name("start_frame")
+                self.show_parameter_by_name("end_frame")
```
SavagePencil: match/case safety net

SavagePencil (changes requested):

griptape_nodes_library/utils/video_utils.py
shhlife: Claude says:

sys.float_info.epsilon exists (≈ 2.22e-16) but it's machine precision — the smallest difference between adjacent floats near 1.0. That's the wrong tool here.

1e-5 is intentional and domain-specific: "is the start within 10 microseconds of the beginning of the video?" It's a tolerance that accounts for floating point results from frame / frame_rate that are nominally zero but might be slightly off. float_info.epsilon would be far too tight — it'd only match values essentially indistinguishable from 0.0 at the machine level, not "close enough to the start that stream copy is safe."

The 1e-5 value is correct. It's not trying to detect "is this exactly zero?" — it's answering the video-domain question "is this effectively the start of the file?"

griptape_nodes_library/video/split_video.py
shhlife: That's a new pattern, but happy to do it!

griptape_nodes_library/video/split_video.py
```diff
@@ -55,64 +47,11 @@ class Segment:
...
-        # Re-encode path (example: H.264 video, copy audio)
-        cmd += ["-c:v", "libx264", "-crf", "18", "-preset", "medium", "-c:a", "aac", "-b:a", "192k"]
-
-    cmd += ["-movflags", "+faststart", str(out_path)]
-    return cmd
+    out_path = Path(outdir) / f"{sanitize_filename(seg.title)}.mp4"
+    return build_video_segment_cmd("ffmpeg", input_path, seg.start_sec, seg.end_sec, str(out_path))
 
 
 class SplitVideo(ControlNode):
```
SavagePencil: should this be a SuccessFailure node?

griptape_nodes_library/video/trim_video.py
```diff
@@ -0,0 +1,263 @@
...
+from griptape_nodes_library.utils.video_utils import (
+    MIN_VIDEO_FILE_SIZE,
+    VIDEO_DURATION_BUFFER,
+    smpte_to_seconds,
+    to_video_artifact,
+    validate_url,
+)
+
+
+class TrimVideo(ControlNode):
```
SavagePencil: same Q as earlier, should this be a SuccessFailure?

griptape_nodes_library/video/trim_video.py
```diff
@@ -0,0 +1,263 @@
...
+
+        dest = self._output_file.build_file()
+        saved = dest.write_bytes(video_bytes)
+        video_artifact = VideoUrlArtifact(saved.location)
+        self.set_parameter_value("video_out", video_artifact)
+        self.publish_update_to_parameter("video_out", video_artifact)
+        self.parameter_output_values["video_out"] = video_artifact
+        self.append_value_to_parameter("logs", f"Saved trimmed video: {saved.name}\n")
+
+    def process(self) -> AsyncResult[None]:
```
SavagePencil: can we use `aprocess` to simplify the signature

griptape_nodes_library/video/split_video.py
SavagePencil: maybe make this `aprocess` to simplify?

griptape_nodes_library/video/split_video.py
shhlife: ah, that's a good point - let me check and see what would happen with migration..

SavagePencil: > ah, that's a good point - let me check and see what would happen with migration..

They'll migrate seamlessly. The Success path uses the same name as the existing Control Out. You'll have to add the error handling bits, tho.

shhlife: yep! they do - I just tested :)

SavagePencil (approved):

PR #288, opened 2026-05-29
feat: add sequence-scanning nodes (Scan / Scan Split / Inspect)

### LINKED ISSUE #4712: ScanSequenceRequest should return macro paths, not absolute paths
## Problem

`ScanSequenceRequest` currently returns absolute paths for the files it finds. For consistency with the rest of the system, which uses macro paths (e.g. `{inputs}/...`), the scan should return macro-form paths instead.

For example, scanning for `{inputs}/output_###.exr` should return:
- `{inputs}/output_001.exr`
- `{inputs}/output_002.exr`
- `{inputs}/output_003.exr`

rather than the resolved absolute paths.

## Origin

Raised in a team design discussion about how sequence scans should represent their results. The consensus was that since everything else in the system works in macros, scan results should be returned as macros too.

## Impact

Returning absolute paths is inconsistent with the rest of the system's macro-based path handling, making scan output awkward to compose with other nodes that expect macro paths.

## Notes

A complementary utility is being tracked separately to resolve macro paths to absolute paths on demand (for cases where a user needs to see where files actually sit on disk).

SavagePencil:
> 🚨 **DO NOT MERGE BEFORE THE ENGINE RELEASE LANDS** 🚨
>
> This PR depends on engine PR https://github.com/griptape-ai/griptape-nodes/pull/4771 (which closes [griptape-nodes#4712](https://github.com/griptape-ai/griptape-nodes/issues/4712)). The library calls `ScanSequencesRequest(path=..., no_token_behavior=...)` — both of those are new fields on the engine's request payload. Until the engine wheel containing those changes is **released and pinned** in this library's `pyproject.toml`, the library will fail at import time on consumer machines.
>
> Concretely: `make check/types` will keep showing pyright errors on `path` / `no_token_behavior` / `directory_had_matching_files` / `discovered_first` / `discovered_last` accesses until the wheel ships. That's the symptom; release order is the cause.

## Summary

Adds three sequence-handling nodes (Scan / Scan Split / Inspect) plus a reusable `AdvancedSequenceControls` component. All are built on top of the engine's `ScanSequencesRequest`, including the new macro-path round-trip and `NoTokenBehavior` dispatch landing in the companion engine PR linked above.

### Nodes

- **Scan Sequence** (`ScanSequenceNode`) — the 80% artist entry point. Takes a path or pattern (`{inputs}/render.####.png`, `/work/render.%04d.png`, or even a single literal file). Top-level outputs are `paths: list[str]` and `sequence: Sequence` — both visible by default, both keep macro-form when a macro was supplied. Default policy is `SKIP` (gaps drop out of the entries list).
- **Scan Split Sequence** (`ScanSplitSequenceNode`) — multi-sequence variant. Always uses the engine's `SPLIT` policy, producing `sequences: list[Sequence]` — one per contiguous run of present items. The `sequences` tooltip points users at **GetFromList** / **ForEach Group** for routing one sub-sequence into Inspect Sequence (workaround for [Display List type-narrowing limitation #298](https://github.com/griptape-ai/griptape-nodes-library-standard/issues/298)).
- **Inspect Sequence** (`InspectSequenceNode`) — pass-through inspection. Outputs a multi-line summary, the sequence's existing entries, gaps, gap count, entry count, and `is_complete`. Reactive: outputs recompute on connection / value change without a node run via `after_value_set` / `after_incoming_connection*` hooks.

### Shared component: `AdvancedSequenceControls`

`advanced_sequence_component.py` ships the `AdvancedSequenceControls` class (renamed from the older `ItemRangeGroup` placeholder) and the `resolve_bounds` / `behavior_for_request` helpers. Any future node that consumes a sequence path can drop this into its `__init__`, add the resulting `ParameterGroup` to the node, and forward `after_value_set` to the component — the entire collapsed *Advanced Sequence Control* group comes along automatically.

### What's in the Advanced group

- **When there's no sequence marker (e.g., ###)** — drives the engine's `NoTokenBehavior` enum. Default *Treat as a single file* so `render.0002.png` returns one file, not the surrounding sequence. Other options: *Treat as part of a sequence* (walks siblings) and *Fail unless a token is present* (strict-mode failure for pipelines).
- **Start at** / **Start at item** / **Skip items from start** — three-mode item-range start. Mode dropdown drives visibility of the value sub-parameter.
- **End at** / **End at item** / **Skip items from end** — mirror.
- All three dropdowns carry `help` badges with worked examples (consistent with the *What to do at gaps* badge on `missing_item_policy`).

### Behavior worth flagging

- **Empty results emit a real (empty) `Sequence`, not `None`.** Inspect Sequence and other downstream nodes treat `None` as "no sequence connected" and short-circuit; emitting a populated-but-empty Sequence (with `first=0, last=-1`, empty `entries`, the actual policy) lets them render a meaningful "0 entries" view instead.
- **Macro paths round-trip end-to-end.** Wire `{inputs}/render.####.png` into Scan Sequence and you get back `paths` and `sequence.entries[*].path` that still say `{inputs}` — workflows stay portable across collaborators whose `{inputs}` resolves differently.
- **`fail_on_empty_result: bool = True`** with diagnostic-aware status strings built off the engine's empty-result diagnostics (`directory_had_matching_files`, `discovered_first/last`). Friendly label: *Fail when no items are found*. Off → empty Sequence + opt-out status, useful for sweeps that legitimately may find nothing.
- **`NoTokenBehavior`** routed through the friendly dropdown described above. Default `SINGLE_FILE` so a literal numbered filename returns just that one file.
- **Friendly display names** — `template` → *Path or pattern*; `missing_item_policy` → *What to do at gaps*; `fail_on_empty_result` → *Fail when no items are found*; `no_token_behavior` → *When there's no sequence marker (e.g., ###)*.
- **Tooltips and badges** follow strict tone rules from the review pass: no programmer jargon, no internal enum names, no prescriptive "most artists want this" claims. Every Advanced-group dropdown carries a `help` badge alongside its tooltip.

## Test plan

- [ ] **Scan Sequence (happy path, macro)**: point at `{inputs}/sequences/01_contiguous/render.####.png`. Verify `paths` populates with five `{inputs}/...` entries (macro head intact) and the `sequence` output carries the structured Sequence.
- [ ] **Scan Sequence (empty, default)**: scan a directory that doesn't match → fail-on-empty fires with a diagnostic status string ("No items matched…" / padding-mismatch / subset-clipped).
- [ ] **Scan Sequence (empty, opt-in)**: flip *Fail when no items are found* off → node succeeds, `sequence` is the empty Sequence (not `None`), status notes the opt-out.
- [ ] **Token-less path, default**: `render.0002.png` against a directory holding `render.0001..0005.png` returns a 1-item sequence containing just `render.0002.png` (regression pin).
- [ ] **Token-less path, explore**: same fixture, dropdown set to *Treat as part of a sequence* → returns all five sibling renders.
- [ ] **Token-less path, reject**: same fixture, dropdown set to *Fail unless a token is present* → fails with an INVALID_TEMPLATE message asking for a marker.
- [ ] **Scan Split Sequence**: scan a directory with three contiguous runs (e.g. `[1..3, 5..7, 10]`) → produces three `Sequence` objects.
- [ ] **Inspect Sequence (reactive)**: connect to either Scan node's output → outputs (summary, entry counts, gap list, is_complete) populate without a node run. Disconnect → outputs reset.
- [ ] **Inspect Sequence (empty)**: connect to a failed Scan node → renders "0 entries" cleanly instead of "No sequence connected."
- [ ] **Item range controls**: exercise *Specify start* / *Relative to sequence start* / *Relative to sequence end* and verify dropdown-driven visibility toggling and bound resolution.
- [ ] **Advanced badges**: confirm `help` badges render on *When there's no sequence marker*, *Start at*, *End at*.

## Companion engine PR

Engine companion: https://github.com/griptape-ai/griptape-nodes/pull/4771

## Related

- Closes nothing on the library side directly. The work was driven by [griptape-nodes#4712](https://github.com/griptape-ai/griptape-nodes/issues/4712) and ad-hoc UX feedback.
- References [#298](https://github.com/griptape-ai/griptape-nodes-library-standard/issues/298) — Display List's `_determine_item_type` falls back to `"any"` for non-curated types, breaking some downstream type-narrowed connections from `list[Sequence]`. Workaround in this PR: tooltips on Split's `sequences` output point users at GetFromList / ForEach Group instead.

### DIFF
--- a/griptape_nodes_library/sequence/inspect_sequence.py
+++ b/griptape_nodes_library/sequence/inspect_sequence.py
@@ -1,0 +1,219 @@
+"""InspectSequenceNode: surface a Sequence's structure for expert/technical workflows.
+
+Pure pass-through inspection node. The `sequence` parameter is both INPUT and
+OUTPUT — the same Sequence flows through unchanged — plus a handful of
+structural primitives that downstream expert workflows commonly need:
+
+- `summary: str` (multiline) — human-readable description of the Sequence,
+  suitable for a Display node or a Note. Surfaces range, policy, padding,
+  pattern, directory, gap count, dropped-negatives count, and a peek at the
+  first/last entry paths.
+- `existing_entries: list[SequenceEntry]` — only the entries whose number is
+  actually present on disk (filters out FILL_NEAREST's gap-fill duplicates).
+- `gaps: list[int]` — sorted item numbers absent from the active range.
+- `is_complete: bool` — True if there are zero gaps in the active range.
+- `gap_count: int` — `len(seq.missing_numbers)`.
+- `entry_count: int` — `len(seq.entries)`.
+
+The Sequence pass-through means an artist can drop this node into the middle
+of a chain without breaking it: input → InspectSequence → downstream Sequence
+consumer.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from griptape_nodes.common.sequences import Sequence
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import BaseNode, SuccessFailureNode
+from griptape_nodes.exe_types.param_types.parameter_int import ParameterInt
+from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
+
+
+class InspectSequenceNode(SuccessFailureNode):
+    """Inspect a Sequence and surface its structural primitives.
+
+    On total failure (no input connected, or unexpected input shape), emits
+    empty/default outputs and routes through the Failure control-flow edge.
+    """
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+
+        self._sequence_param = Parameter(
+            name="sequence",
+            tooltip=(
+                "The Sequence to inspect. Pass-through: the same Sequence flows through unchanged "
+                "as the output, so this node can sit inline between any two Sequence-aware nodes."
+            ),
+            type="Sequence",
+            input_types=["Sequence"],
+            output_type="Sequence",
+            default_value=None,
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT},
+        )
+        self.add_parameter(self._sequence_param)
+
+        self._summary_param = ParameterString(
+            name="summary",
+            default_value="",
+            tooltip=(
+                "Human-readable description of the Sequence: range, policy, padding, "
+                "pattern, directory, gap count, and a peek at the first/last entries."
+            ),
+            allowed_modes={ParameterMode.OUTPUT},
+            multiline=True,
+        )
+        self.add_parameter(self._summary_param)
+
+        self._existing_entries_param = Parameter(
+            name="existing_entries",
+            tooltip=(
+                "List of SequenceEntry objects whose number is actually present on disk. "
+                "Under FILL_NEAREST, this filters out gap entries (which carry the nearest "
+                "neighbor's path); under SKIP / ABORT it equals `sequence.entries`."
+            ),
+            type="list",
+            output_type="list[SequenceEntry]",
+            default_value=[],
+            allowed_modes={ParameterMode.OUTPUT},
+        )
+        self.add_parameter(self._existing_entries_param)
+
+        self._gaps_param = Parameter(
+            name="gaps",
+            tooltip="Sorted list of item numbers missing from the active range.",
+            type="list",
+            output_type="list[int]",
+            default_value=[],
+            allowed_modes={ParameterMode.OUTPUT},
+        )
+        self.add_parameter(self._gaps_param)
+
+        self._is_complete_param = Parameter(
+            name="is_complete",
+            tooltip="True when the Sequence has no gaps in its active range.",
+            type="bool",
+            output_type="bool",
+            default_value=False,
+            allowed_modes={ParameterMode.OUTPUT},
+        )
+        self.add_parameter(self._is_complete_param)
+
+        self._gap_count_param = ParameterInt(
+            name="gap_count",
+            default_value=0,
+            tooltip="Number of items missing from the active range (`len(seq.missing_numbers)`).",
+            allowed_modes={ParameterMode.OUTPUT},
+        )
+        self.add_parameter(self._gap_count_param)
+
+        self._entry_count_param = ParameterInt(
+            name="entry_count",
+            default_value=0,
+            tooltip="Number of entries in the Sequence (`len(seq.entries)`).",
+            allowed_modes={ParameterMode.OUTPUT},
+        )
+        self.add_parameter(self._entry_count_param)
+
+        self._create_status_parameters(
+            result_details_tooltip="Details about the Sequence inspection",
+            result_details_placeholder="Inspection results will appear here.",
+        )
+
+    def validate_before_workflow_run(self) -> list[Exception] | None:
+        """Skip SuccessFailureNode's status reset.
+
+        The parent class clears `was_successful` / `result_details` here, which
+        causes downstream re-resolutions of this node to leave the status reading
+        False / placeholder if the engine doesn't re-run `process()` immediately
+        afterward. We skip the reset and let `process()` own the status entirely.
+        """
+        return BaseNode.validate_before_workflow_run(self)
+
+    def validate_before_node_run(self) -> list[Exception] | None:
+        """Skip SuccessFailureNode's status reset (see `validate_before_workflow_run`)."""
+        return BaseNode.validate_before_node_run(self)
+
+    def process(self) -> None:
+        self._clear_execution_status()
+
+        raw = self.get_parameter_value(self._sequence_param.name)
+        if raw is None:
+            self._emit_failure("No sequence connected.")
+            return
+
+        # Accept either a Sequence instance (typical) or a dict (e.g. after a
+        # save/load round-trip if the engine handed it back as JSON).
+        sequence = Sequence.model_validate(raw)
+
+        gaps = sorted(sequence.missing_numbers)
+        gap_count = len(gaps)
+        entry_count = len(sequence.entries)
+        is_complete = gap_count == 0
+
+        existing_entries = [entry for entry in sequence.entries if entry.number in sequence.present_numbers]
+
+        self.parameter_output_values[self._sequence_param.name] = sequence
+        self.parameter_output_values[self._summary_param.name] = _build_summary(sequence)
+        self.parameter_output_values[self._existing_entries_param.name] = existing_entries
+        self.parameter_output_values[self._gaps_param.name] = gaps
+        self.parameter_output_values[self._is_complete_param.name] = is_complete
+        self.parameter_output_values[self._gap_count_param.name] = gap_count
+        self.parameter_output_values[self._entry_count_param.name] = entry_count
+
+        details = (
+            f"Inspected sequence: items {sequence.first}..{sequence.last} "
+            f"({entry_count} entries, {gap_count} gap(s), policy={sequence.policy.value})."
+        )
+        self._set_status_results(was_successful=True, result_details=details)
+
+    def _emit_failure(self, details: str) -> None:
+        """Emit empty/default outputs and mark the node failed."""
+        self.parameter_output_values[self._sequence_param.name] = None
+        self.parameter_output_values[self._summary_param.name] = ""
+        self.parameter_output_values[self._existing_entries_param.name] = []
+        self.parameter_output_values[self._gaps_param.name] = []
+        self.parameter_output_values[self._is_complete_param.name] = False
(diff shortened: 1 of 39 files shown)


griptape_nodes_library/sequence/scan_sequence.py
```diff
@@ -0,0 +1,352 @@
...
+
+    On total failure (bad template, unresolvable project vars, missing
+    directory, no matches, ABORT-policy gap), the node emits empty outputs
+    and routes through the Failure control-flow edge.
+    """
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+
+        self._template_param = ParameterString(
```
shhlife: It'd be great if this could be a file path so we can choose a path on disk.. in Nuke for example, when reading images you see sequences displayed with the seqence information:

Image

This makes it much easier to select sequence paths on disk.

griptape_nodes_library/sequence/scan_sequence.py
```diff
@@ -0,0 +1,352 @@
...
+            output_type="list[str]",
+            default_value=[],
+            allowed_modes={ParameterMode.OUTPUT},
+        )
+        self.add_parameter(self._paths_param)
+
+        with ParameterGroup(name="Advanced Sequence Control", collapsed=True) as advanced_group:
+            self._item_range = ItemRangeGroup()
+
+            self._sequence_param = Parameter(
```
shhlife: should this sequence object be in advanced? seems like it'd be handy to have at all times..

griptape_nodes_library/sequence/inspect_sequence.py
```diff
@@ -0,0 +1,219 @@
...
+    """Inspect a Sequence and surface its structural primitives.
+
+    On total failure (no input connected, or unexpected input shape), emits
+    empty/default outputs and routes through the Failure control-flow edge.
+    """
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+
+        self._sequence_param = Parameter(
```
shhlife, line 55: I wonder if the data can update based on connection instead of having to run the node? is it scanning things or is it just looking at the sequence object?

griptape_nodes_library/sequence/scan_sequence.py
```diff
@@ -0,0 +1,352 @@
...
+            tooltip=(
+                "Macro template. May contain a fileseq sequence token (`####` or `%04d`). "
+                "Non-sequence templates are NOT supported here."
+            ),
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            ui_options={"display_name": "Sequence path"},
+        )
+        self.add_parameter(self._template_param)
+
+        self._policy_param = ParameterString(
```
shhlife, line 87: Can we add an info badge to this that explains how each of these work with examples?

Badges can use markdown to make things cleaner - there's a claude skill for it

griptape_nodes_library/sequence/scan_split_sequence.py
```diff
@@ -0,0 +1,235 @@
...
+            tooltip=(
+                "Macro template. May contain a fileseq sequence token (`####` or `%04d`). "
+                "Non-sequence templates are NOT supported here."
+            ),
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            ui_options={"display_name": "Sequence path"},
+        )
+        self.add_parameter(self._template_param)
+
+        self._sequences_param = Parameter(
```
shhlife, line 67: I tried to use this to split up a sequence, but couldn't inspect the split sequences?

Image

the error was
```
No sequence connected.
```

shhlife (changes requested):
A couple of notes..

I was confused at one point because I was looking at images here:

{outputs}/Adjust Image EQ_output_##.png

because I had a bunch of images with 21, 22, 23, 24, etc.. and it was returning errors because it didn't recognize those (I also had images with 1, 2, 3, 4). If I changed to:

{outputs}/Adjust Image EQ_output_#.png

it worked..

so if we have something that has "no" padding.. but all the numbers have 2 digits.. should that work?



btw - the inspect sequence node is awesome.

griptape_nodes_library/sequence/scan_sequence.py
SavagePencil: No argument there; I thought since it’s rarer to use except for more advanced stuff I’d hide it away. Maybe a clearer display name to distinguish from “pile of file paths”?

griptape_nodes_library/sequence/inspect_sequence.py
SavagePencil: Yeah totally. This one’s cheap.  Good call

griptape_nodes_library/sequence/scan_sequence.py
SavagePencil: Yes

griptape_nodes_library/sequence/scan_split_sequence.py
SavagePencil: Dang

griptape_nodes_library/sequence/scan_sequence.py
shhlife: I think if you have information in either a badge or in the tooltip that'll suffice?

griptape_nodes_library/sequence/scan_split_sequence.py
SavagePencil: This is actually a failure of Display List, which is doing some dumb type-specific testing in there and Sequence isn't one of them. Confirmed this works with Get From List node; creating a follow-up task.

SavagePencil: https://github.com/griptape-ai/griptape-nodes-library-standard/issues/298

griptape_nodes_library/sequence/scan_split_sequence.py
```diff
@@ -0,0 +1,233 @@
...
+            tooltip=(
+                "A path to a numbered file sequence. Use a token like `####`, `%04d`, or `@@@` "
+                "where the frame number goes — for example `render.####.png` matches "
+                "`render.0001.png`, `render.0002.png`, … A plain path with no token works too; "
+                "it's read as a sequence of one. Project macros like `{inputs}/render.####.png` "
+                "are accepted and preserved end-to-end (the items you get back keep the "
+                "`{inputs}` form), so workflows stay portable across machines."
+            ),
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            ui_options={"display_name": "Path or pattern"},
```
shhlife, line 63: Add a `FileSystemPicker` trait please

```python
            traits={FileSystemPicker(allow_files=True, allow_directories=False)},
```

griptape_nodes_library/sequence/inspect_sequence.py
```diff
@@ -0,0 +1,313 @@
...
+            name="sequence",
```
shhlife, line 56: I found when I connected a split sequence to this I got the following error on connection, but the result _was_ correct.. not sure why it was still getting the error? just need to update input_types?

```
Error: SetParameterValue Failed
Description: "Attempted to set parameter value for 'Inspect Sequence.sequence'. Failed because the value's type of 'any' was not in the Parameter's list of allowed types: ['Sequence']."
Timestamp: 6/10/2026, 9:44:38 AM
```

Image

shhlife (changes requested):
just 2 quick notes

griptape_nodes_library/sequence/inspect_sequence.py
SavagePencil: yeah, Display List is borked: https://github.com/griptape-ai/griptape-nodes-library-standard/issues/298

I'll fix that separately.

griptape_nodes_library/sequence/scan_split_sequence.py
SavagePencil: OK if I do that as a follow-up? I see you already created an issue for allowing selecting a sequence in the File Picker. Here was my concern with adding it now:
1. FilePicker, today, returns a single file path
2. We default in the advanced group, to "if you get a single file path, treat it as a sequence of 1"
3. This might be disorienting to have to go in and change the single file path to "single file path = scan based on it"

shhlife: yeah, I found it frustrating to not have the file picker because I was immediately like "ugh, I don't want to have to go and find a path! how am I going to do that on a remote engine??"

so I wanted the file picker.. does the engine currently give the option of returning the path as a sequence?

SavagePencil: confirming what you'd want here is:
1. Open file picker
2. Select `/Users/Jason/Source Materials/render_0004.png`
3. Click the Use Sequences checkbox in the File picker?
4. Confirm in File Picker
5. Node converts to `{inputs}/render_####.png`

Is that the desired experience? If so, I'll need to do some engine juggling to make that happen, so wondering if I can defer that as a TODO

shhlife (approved):

PR #292, opened 2026-05-30
Update: Crop Image node - replace modal editor with inline crop editor widget

### LINKED ISSUE #291: Crop Image node: replace the modal/slider crop controls with an inline Widget
## Problem

The `CropImage` node (`griptape_nodes_library/image/crop_image.py`) currently exposes cropping through an "Open Crop Editor" `ParameterButton` plus a group of `left` / `top` / `width` / `height` integer sliders. Cropping is a spatial, direct-manipulation task, and adjusting it numerically (or through a separate modal) is clunky compared to dragging a crop rectangle directly on the image.

This issue tracks converting the crop controls to an inline custom **Widget** that renders the image and lets the user draw/drag the crop region in place, keeping the node's `left`/`top`/`width`/`height` parameters in sync.

## Origin

Follow-up to #290 (Crop Image node throwing `'dict' object has no attribute 'value'`). While triaging that bug it came up that the crop UX itself should move to a Widget rather than the current button-plus-sliders approach. This is a UX/enhancement request, not a regression.

## Impact

Not blocking — the node functions today. This is an ergonomics improvement: a direct-manipulation crop widget is faster and less error-prone than entering pixel coordinates by hand, and it brings the node in line with the library's other Widget-based media nodes.

## Notes / prior art

The library already has a Widget pattern to follow:

- `griptape_nodes.traits.widget.Widget` trait (see `griptape_nodes_library/video/adjust_mask_size.py`, `griptape_nodes_library/splat/load_splat.py`)
- JS widgets under `widgets/` and `griptape_nodes_library/*/widgets/` (e.g. `MaskAdjustmentPreview.js`, `SplatViewer.js`)
- The `griptape-nodes-widget-dev` skill in `.claude/skills/` documents widget lifecycle and the Python ↔ JS value bridge

Scope to consider when picking this up:

- Render the `input_image` in the widget and overlay a draggable/resizable crop rectangle.
- Two-way bind the rectangle to the existing `left` / `top` / `width` / `height` parameters so downstream behavior is unchanged.
- Decide the fate of the existing "Open Crop Editor" button and the slider group (replace vs. keep as a fallback).
- Make sure the dict-normalization fix from #290 is in place first, since the widget will consume the same `input_image` value.

shhlife:
https://www.loom.com/share/1e6dbee8955944abb266d8eb872b1673

## Summary

- Replaces the modal crop editor button + numeric sliders with an inline `CropImageEditor` widget that renders the image and lets users drag/resize the crop rectangle directly in the node
- Fixes `'dict' object has no attribute 'value'` error when `input_image` - this was legacy and was also breaking macro paths. Using `_extract_image_path` 
- Hides the `input_image` property row (keeps the input socket) to reduce visual clutter
- Pre-publishes the full `crop_editor` dict before individual param publishes to eliminate the snap-then-jump artifact during drag

## Widget features

- Drag to move, drag handles to resize, drag outside crop to draw a new rect
- Shift-drag for square/locked-aspect-ratio operations
- Rule-of-thirds guide overlay and dashed blue zoom/rotate preview rect
- Sidebar with aspect ratio presets (1:1, 16:9, 4:3, 3:2, 9:16, 2:3, 3:4) — preserves the current crop's shortest dimension when switching ratios
- Reset button returns crop to full image
- Zoom and rotate sliders with reset controls
- Incoming wire connections lock the affected handles and sidebar buttons (read-only visual)
- Lucide icons for reset buttons; Tailwind CSS variables for full light/dark theme support
- Canvas letterboxes within available node height using `flex:1 1 0` + `Math.min(areaW/imgNatW, areaH/imgNatH)` scaling

## Widget file layout

```
widgets/CropImageEditor/
  CropImageEditor.js   ← main entry point
  _styles.js           ← visual constants
  _icons.js            ← Lucide SVG paths + mkIcon factory
  _sidebar.js          ← aspect ratio preset buttons
  _footer.js           ← zoom/rotate sliders + status bar
```

## Closes

Fixes #290
Closes #291

### DIFF
--- a/.claude/skills/griptape-nodes-widget-dev/SKILL.md
+++ b/.claude/skills/griptape-nodes-widget-dev/SKILL.md
@@ -106,6 +106,112 @@ Do **not** call `set_parameter_value` on the widget param from `__init__` — un
 
 ---
 
+## Preventing sync loops between Python and the widget
+
+When the widget emits a change and Python writes individual parameters back, `after_value_set` fires for each one. Without guards these trigger another widget push, which triggers another emit, and so on.
+
+Use two flags — one for each direction — to break the loop:
+
+```python
+self._syncing_to_widget = False  # Python → widget in progress
+self._syncing_to_params = False  # widget → params in progress
+
+def after_value_set(self, parameter, value):
+    if parameter.name == "my_widget_param" and not self._syncing_to_widget:
+        # widget changed → sync to individual params
+        self._syncing_to_params = True
+        try:
+            self.set_parameter_value("x", value.get("x", 0))
+            self.publish_update_to_parameter("x", value.get("x", 0))
+        finally:
+            self._syncing_to_params = False
+
+    elif not self._syncing_to_params and parameter.name == "x":
+        # individual param changed → push to widget
+        self._syncing_to_widget = True
+        try:
+            self._push_widget({...})
+        finally:
+            self._syncing_to_widget = False
+```
+
+---
+
+## Pre-publish the widget dict before individual param publishes
+
+**Problem (snap-then-jump):** When the widget emits and Python calls `publish_update_to_parameter("x", 300)` for each individual param, FlowEditor re-renders *all* widgets using the currently-stored widget dict value. If the stored dict still has the old value (`x: 100`), the widget snaps back to `x: 100` on every individual param publish.
+
+**Fix:** publish the full widget dict first, before any individual param publishes. This ensures the stored value is authoritative before the re-render fires.
+
+```python
+def _sync_params_from_widget(self, widget_dict: dict) -> None:
+    # ✅ pre-publish the widget dict FIRST
+    self.publish_update_to_parameter("my_widget_param", widget_dict)
+
+    self._syncing_to_params = True
+    try:
+        for key, val in widget_dict.items():
+            self.set_parameter_value(key, val)
+            self.publish_update_to_parameter(key, val)
+    finally:
+        self._syncing_to_params = False
+```
+
+---
+
+## Detecting wire connections to lock widget controls
+
+When an upstream node wires into a parameter, the widget should show that parameter as read-only (don't let the user drag a handle that would override a connected value).
+
+```python
+from griptape_nodes.retained_mode.events.connection_events import (
+    ListConnectionsForNodeRequest,
+    ListConnectionsForNodeResultSuccess,
+)
+
+def _get_locked_params(self) -> list[str]:
+    watched = {"x", "y", "width", "height"}
+    try:
+        result = GriptapeNodes.handle_request(ListConnectionsForNodeRequest(node_name=self.name))
+        if isinstance(result, ListConnectionsForNodeResultSuccess):
+            return [c.target_parameter_name for c in result.incoming_connections
+                    if c.target_parameter_name in watched]
+    except Exception:
+        pass
+    return []
+
+def after_incoming_connection(self, source_node, source_parameter, target_parameter):
+    if target_parameter.name in self._watched_params:
+        self._push_widget({**self._current_widget_dict(), "locked": self._get_locked_params()})
+    return super().after_incoming_connection(source_node, source_parameter, target_parameter)
+
+def after_incoming_connection_removed(self, source_node, source_parameter, target_parameter):
+    if target_parameter.name in self._watched_params:
+        self._push_widget({**self._current_widget_dict(), "locked": self._get_locked_params()})
+    return super().after_incoming_connection_removed(source_node, source_parameter, target_parameter)
+```
+
+In the widget JS, read `value.locked` and use it to disable the relevant interactive elements.
+
+---
+
+## Hiding a parameter's property row (keep socket only)
+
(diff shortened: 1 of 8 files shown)


.claude/skills/griptape-nodes-widget-dev/SKILL.md
```diff
@@ -106,6 +106,112 @@ Do **not** call `set_parameter_value` on the widget param from `__init__` — un
 
 ---
 
+## Preventing sync loops between Python and the widget
```
cjkindel, line 109: Widgets are JS, confused on what value the Python examples here bring to Widget authoring? Seems like more valuable for engine dev in support of widgets?

griptape_nodes_library/image/crop_image.py
```diff
@@ -76,25 +78,40 @@ def __init__(self, **kwargs) -> None:
         self.MAX_WIDTH = MAX_WIDTH
         self.MAX_HEIGHT = MAX_HEIGHT
         self._processing = False  # Lock to prevent live cropping during process()
```
cjkindel: We don't use _processing anymore (b/c live cropping was removed)

griptape_nodes_library/image/crop_image.py
```diff
@@ -524,18 +698,8 @@ def _validate_image(self) -> list[Exception]:
             exceptions.append(Exception(msg))
             return exceptions
 
-        # Validate input artifact type
-        if isinstance(input_artifact, dict):
-            # Convert dict to ImageUrlArtifact for validation
-            try:
-                input_artifact = dict_to_image_url_artifact(input_artifact)
```
cjkindel, line 531: PR says we normalize input via dict_to_image_url_artifact but now we have removed that fn call?

griptape_nodes_library/image/crop_image.py
```diff
@@ -178,33 +195,201 @@ def __init__(self, **kwargs) -> None:
...
+        try:
+            result = GriptapeNodes.handle_request(ListConnectionsForNodeRequest(node_name=self.name))
+            if isinstance(result, ListConnectionsForNodeResultSuccess):
+                return [
+                    conn.target_parameter_name
+                    for conn in result.incoming_connections
+                    if conn.target_parameter_name in crop_params
+                ]
+        except Exception:
+            pass
```
cjkindel: Why passing here and _resolve_image_url?

griptape_nodes_library/image/crop_image.py
```diff
@@ -178,33 +195,201 @@ def __init__(self, **kwargs) -> None:
...
+        """Build the crop_editor dict from current parameter values."""
+        return {
+            "image_url": image_url,
+            "img_width": img_width,
+            "img_height": img_height,
+            "left": self.get_parameter_value("left") or 0,
+            "top": self.get_parameter_value("top") or 0,
+            "width": self.get_parameter_value("width") or 0,
+            "height": self.get_parameter_value("height") or 0,
+            "zoom": self.get_parameter_value("zoom") or NO_ZOOM,
```
cjkindel: Why 0? Should we use the default values instead?

cjkindel (changes requested):
If removing live-crop intentionally, we should document that in the PR title as it is a change that customers may notice

.claude/skills/griptape-nodes-widget-dev/SKILL.md
shhlife: These are important because some of the changes on the parameters that are _not_ part of the widget affect how the widget behaves - and the widget's behavior affects those parameters.

For example, on this particular widget you can see the left, top, width, and height parameters available to the user

image

Changing the widget affects those parameters, and changing those parameters affects the widget. Without the Python code, these won't work.

cjkindel: Got it, thanks. This is actually a similar problem to workers w/ event loops. Eventually we may want to implement something into Widgets themselves to avoid cycles but this works

griptape_nodes_library/image/crop_image.py
shhlife: ah, yes - `dict_to_image_url_artifact` was removed because it broke macro path inputs — `_extract_image_path` handles the needed cases without it. I'll update the PR.

griptape_nodes_library/image/crop_image.py
shhlife: They're intentional fallbacks.. if `_get_locked_params` fails, it returns just the default - it's not really a blocker. But having `pass` makes debugging harder. Will add warnings.

griptape_nodes_library/image/crop_image.py
shhlife: good catch, adding constants

cjkindel (approved):

.claude/skills/griptape-nodes-widget-dev/SKILL.md
shhlife: totally! I have something to show at CX that is also relevant in this space. There's a bunch of widget stuff we can do as boostrap/placeholder that'll make consistent widget interaction better. :)

PR #338, opened 2026-06-17
feat: declare node models in the model catalog

collindutter:
Supersedes #316.

Declares the models the library's nodes use in a `model_catalog`, as descriptive metadata rather than something nodes read at runtime.

Bumps `library_schema_version` to `0.10.0` and declares one `model_catalog` under a single `griptape_cloud` provider covering every node that calls the Griptape Cloud proxy: the chat models (`Agent`, `GriptapeCloudPrompt`) and the image, video, audio, 3D, and splat proxy nodes. Each node references the models it uses with `model_usage`, and each model carries its `provider_model_id`, family, and key support, so policy, admin, and UI consumers can resolve what every node uses.

The catalog is the contract the library exposes to the platform; the node Python is the implementation that satisfies it. Nodes do not read the catalog at runtime, so a wrong or missing catalog never changes node behavior and nodes stay self-contained. A unit test resolves the chat nodes' `model_usage` against the catalog and asserts it matches the model list they serve, so the manifest and the code cannot drift. An earlier revision had the chat nodes build their dropdowns from `get_declared_models(self)`; that inverted the dependency and made the catalog load-bearing, so it was backed out in favor of the test.

`provider_model_id` records the proxy model id with any trailing `:task` routing suffix stripped, so sibling nodes that drive the same upstream model (the `LTX`, `Kling`, and `Wan` families) share one entry; the two standalone Kling task nodes keep their literal endpoint ids since they expose no base-model selector. In-payload selectors (Tripo version, Topaz operation presets, OpenAI background) stay parameters, so the catalog entry is whatever changes the proxy URL. Models whose vendor exposes a `ProxyApiKeyProviderConfig` are marked `SUPPORTS_CUSTOMER_KEY_OR_GRIPTAPE_KEY`; Google (Nano Banana image and Veo) stays `REQUIRES_GRIPTAPE_KEY` because both its nodes sit in the BYOK disabled set.

Requires a released engine that understands the `0.10.0` catalog schema (https://github.com/griptape-ai/griptape-nodes-engine/pull/4838) so the declarations parse at library load.

Toward https://github.com/griptape-ai/griptape-nodes-engine/issues/4805.

### DIFF
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -56,402 +56,6 @@
       {
         "type": "model_catalog",
         "providers": {
-          "anthropic": {
-            "display_name": "Anthropic",
-            "terms_url": "https://www.anthropic.com/legal/commercial-terms",
-            "notes": "Direct API access. Requires ANTHROPIC_API_KEY (BYOK).",
-            "models": {
-              "anthropic_claude_opus_4_7": {
-                "display_name": "Claude Opus 4.7",
-                "family": "Claude 4.7",
-                "provider_model_id": "claude-opus-4-7",
-                "key_support": "REQUIRES_CUSTOMER_KEY"
-              },
-              "anthropic_claude_sonnet_4_6": {
-                "display_name": "Claude Sonnet 4.6",
-                "family": "Claude 4.6",
-                "provider_model_id": "claude-sonnet-4-6",
-                "key_support": "REQUIRES_CUSTOMER_KEY"
-              },
-              "anthropic_claude_haiku_4_5": {
-                "display_name": "Claude Haiku 4.5",
-                "family": "Claude 4.5",
-                "provider_model_id": "claude-haiku-4-5",
-                "key_support": "REQUIRES_CUSTOMER_KEY"
-              }
-            }
-          },
-          "openai": {
-            "display_name": "OpenAI",
-            "terms_url": "https://openai.com/policies/terms-of-use",
-            "notes": "Direct API access. Requires OPENAI_API_KEY (BYOK).",
-            "models": {
-              "openai_gpt_5_2": {
-                "display_name": "GPT-5.2",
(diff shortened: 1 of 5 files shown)


griptape_nodes_library.json
```diff
@@ -2575,6 +3286,18 @@
...
+        ],
+        "declarations": [
+          {
+            "type": "model_usage",
+            "model_ids": [
+              "grok_3_beta",
+              "grok_3_fast_beta",
+              "grok_3_mini_beta",
+              "grok_3_mini_fast_beta",
+              "grok_2_vision_1212"
```
cjkindel: How can we de-dupe these model_id lists so the nodes don't contain them too? Also not confident these are 100% accurate, grok_2_vision_1212 is deprecated and I don't see [grok-imagen-image](https://github.com/griptape-ai/griptape-nodes-library-standard/blob/62947656e376b8da962c2ecd04a754d5468d08c8/griptape_nodes_library/image/grok_image_edit.py#L45)?

cjkindel: Wait, are these just for text/chat models? Guess that answers 1/2 my Q

collindutter: Good question, let me put some thought into that..

collindutter: Added a method to get the libraries declarations https://github.com/griptape-ai/griptape-nodes-engine/pull/4842

cjkindel (approved):

griptape_nodes_library.json
collindutter: @cjkindel I ended up reverting the node changes that consume the catalog. Each proxy is slightly different and I think we need to have a discussion with @SavagePencil on whether we want to mix these two concerns (declaring/consuming).

The catalog should be updated to include all models though.

PR #341, opened 2026-06-17
Fix: Collapse input media thumbnails on proxy nodes to reduce node height

shhlife:
## Summary

Closes #334

- Set `hide_property=True` on all input `ParameterImage` and `ParameterVideo` parameters across proxy image and video generation nodes, collapsing the large asset thumbnail widget while preserving the input connector
- Set `allowed_modes={ParameterMode.INPUT}` (removing `PROPERTY`) on these same parameters so the drop-zone widget doesn't appear in the properties panel either
- Applied to 28 files: all proxy image nodes (`flux`, `grok`, `seedream`, `qwen`, `openai`, `google`, `wan`) and video nodes (`kling`, `sora`, `seedance`, `veo3`, `grok`, `minimax hailuo`, `omnihuman`, `ltx`, `wan`, and more)

## Test plan

- [ ] Open affected proxy nodes in the canvas and confirm input image/video connectors are still present
- [ ] Confirm no large thumbnail drop-zone widget appears on the node canvas for input media parameters
- [ ] Confirm the node height is visually reduced compared to before
- [ ] Wire an upstream image/video node to a proxy node and confirm the connection works correctly
- [ ] Verify output image/video parameters still display their thumbnails as expected


griptape_nodes_library/image/flux_image_generation.py
```diff
@@ -109,13 +109,11 @@ def __init__(self, **kwargs: Any) -> None:
...
+            ParameterImage(
                 name="input_image",
-                input_types=["ImageArtifact", "ImageUrlArtifact", "str"],
-                type="ImageArtifact",
-                default_value=None,
                 tooltip="Optional input image for image-to-image generation (supports up to 20MB or 20 megapixels)",
                 allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+                hide_property=True,
                 ui_options={"display_name": "Input Image"},
             )
```
collindutter, line 118: Why have you done `hide_property=True` in some places and `ui_options.hide_property: True` in others?

shhlife: ParameterList and Parameter don't have the `hide_property` attribute yet - I could add it to the engine, but that'd be a separate task. I can add it, and then we can update these?

collindutter: Got it, it's fine as is. I think we need to take a pass over all these properties in general. Lots of ways to do the same thing..

collindutter (approved):

griptape_nodes_library/image/flux_image_generation.py
shhlife: @collindutter you had me at "we need to take a pass over all these properties in general." :)

PR #343, opened 2026-06-18
Add Seedance provider reference assets

zachgiordano:
## Summary

Adds support for registering media as provider private assets for the **Seedance 2.0** node, ported from the byteplus-proxy library.

- **New node `SeedanceHumanReferenceAsset`** (`video/seedance` category): packages an image, video, or audio as a provider-asset reference for Seedance 2.0 human-reference inputs.
- **`Seedance20VideoGeneration`** now accepts those references on its reference inputs and registers them as private assets at generation time (`asset://{id}`).

Closes https://github.com/griptape-ai/griptape-nodes-library-byteplus-proxy/issues/2

## Gating

Private-asset functionality is active **only** when:
- the model is **Seedance 2.0** (not 2.0 Fast), and
- the node uses **Griptape auth** (not BYOK) — provider assets are registered through the Griptape Cloud proxy on the org's behalf.

Connecting a reference when either condition is unmet raises a clear, specific validation error. Normal (non-asset) media paths are unchanged.

## Access check

The asset node checks org entitlement by probing `GET proxy/v2/assets/<id>` and classifies the result:
- **Granted** (reached the handler) → no error.
- **Denied** (HTTP 403 entitlement gate) → error asking an admin to request access from Foundry; blocks the run.
- **Indeterminate** (401 / 5xx / network) → surfaces the real cause; does **not** claim no-access or block.

## Tests

New unit coverage for the model/BYOK gating and the access-check classification (granted / denied / indeterminate). Existing behavior preserved.

provider_access_required

seedance_human_references

### DIFF
--- a/griptape_nodes_library/video/seedance_2_0_video_generation.py
+++ b/griptape_nodes_library/video/seedance_2_0_video_generation.py
@@ -459,8 +459,10 @@ class Seedance20VideoGeneration(GriptapeProxyNode):
             self._public_reference_video_parameter_1.delete_uploaded_artifact()
             self._public_reference_video_parameter_2.delete_uploaded_artifact()
             self._public_reference_video_parameter_3.delete_uploaded_artifact()
-            # The backend deletes the provider asset on terminal generation state, but the
-            # transient GTC static-storage upload made to feed CreateProviderAsset is ours
+            # Provider assets are reclaimed by the backend: a submitted generation deletes its
+            # linked assets on terminal state, and assets we register but never submit (e.g. a
+            # build failure after registration) are reclaimed by the backend's orphan sweeper.
+            # The transient GTC static-storage upload made to feed CreateProviderAsset is ours
             # to clean up, along with the scratch parameter created to perform the upload (its
             # name is unique per call, so leaving it would accumulate parameters on the node).
             for helper, scratch_name in self._pending_asset_uploads:
@@ -921,7 +923,12 @@ class Seedance20VideoGeneration(GriptapeProxyNode):
         return await self._poll_provider_asset(str(provider_asset_id), headers)
 
     async def _poll_provider_asset(self, provider_asset_id: str, headers: dict[str, str]) -> str:
-        """Poll GET proxy/v2/assets/<id> until ACTIVE; return the provider asset id."""
+        """Poll GET proxy/v2/assets/<id> until ACTIVE; return the provider asset id.
+
+        Transient errors (a network blip, a 5xx, or the eventual-consistency 404 right after the
+        asset id is minted) are logged and retried until attempts are exhausted, mirroring the
+        base class's generation poll. Only a terminal status (FAILED/DELETED) fails immediately.
+        """
         get_url = urljoin(self._proxy_base, f"assets/{provider_asset_id}")
         async with httpx.AsyncClient() as client:
             for attempt in range(ASSET_MAX_ATTEMPTS):
@@ -930,8 +937,12 @@ class Seedance20VideoGeneration(GriptapeProxyNode):
                     response.raise_for_status()
                     result_json = response.json()
                 except Exception as e:
-                    msg = f"{self.name}: failed to poll private asset {provider_asset_id}: {e}"
-                    raise RuntimeError(msg) from e
+                    # Transient — log and retry rather than aborting an otherwise-successful run.
+                    self._log(
+                        f"{self.name} error polling private asset {provider_asset_id} (attempt {attempt + 1}): {e}"
+                    )
+                    await asyncio.sleep(ASSET_POLL_INTERVAL)
+                    continue
 
                 status = result_json.get("status", "unknown")
                 self._log(f"{self.name} private asset {provider_asset_id} status: {status} (attempt {attempt + 1})")
--- a/griptape_nodes_library/video/seedance_2_0_video_generation.py
+++ b/griptape_nodes_library/video/seedance_2_0_video_generation.py
@@ -1,9 +1,14 @@
 from __future__ import annotations
 
+import asyncio
 import json as _json
 import logging
+from contextlib import suppress
 from typing import Any, ClassVar
+from urllib.parse import urljoin
+from uuid import uuid4
 
+import httpx
 from griptape.artifacts import AudioArtifact, ImageArtifact, ImageUrlArtifact
 from griptape.artifacts.audio_url_artifact import AudioUrlArtifact
 from griptape.artifacts.video_url_artifact import VideoUrlArtifact
@@ -23,6 +28,15 @@ from griptape_nodes.files.file import File, FileLoadError
 from griptape_nodes.traits.options import Options
 from griptape_nodes.utils.artifact_normalization import normalize_artifact_input, normalize_artifact_list
 
+from griptape_nodes_library.assets import (
+    ASSET_KIND_AUDIO,
+    ASSET_KIND_IMAGE,
+    ASSET_KIND_VIDEO,
+    ASSET_REFERENCE_TYPE_NAMES,
+    get_provider_asset_kind,
+    get_provider_asset_value,
+    is_provider_asset_reference,
+)
 from griptape_nodes_library.media import coerce_media_url_or_data_uri
 from griptape_nodes_library.proxy import GriptapeProxyNode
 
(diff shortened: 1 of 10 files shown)


griptape_nodes_library/video/seedance_2_0_video_generation.py
```diff
@@ -664,6 +814,148 @@ async def _add_media_inputs_async(self, content_list: list[dict[str, Any]], para
...
+        get_url = urljoin(self._proxy_base, f"assets/{provider_asset_id}")
+        async with httpx.AsyncClient() as client:
+            for attempt in range(ASSET_MAX_ATTEMPTS):
+                try:
+                    response = await client.get(get_url, headers=headers, timeout=60)
+                    response.raise_for_status()
+                    result_json = response.json()
+                except Exception as e:
+                    msg = f"{self.name}: failed to poll private asset {provider_asset_id}: {e}"
+                    raise RuntimeError(msg) from e
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding against the current head of the branch.

**Finding: provider-asset polling aborts on the first transient error.**

Any single exception in a poll iteration (a `raise_for_status()` on a transient 5xx, a network blip, or a 404 during the eventual-consistency window right after `_create_provider_asset` returns the new id) immediately raises and aborts the whole generation. Over the roughly three-minute poll window (`ASSET_MAX_ATTEMPTS * ASSET_POLL_INTERVAL`) a single hiccup is plausible.

This diverges from the resilience pattern the base class already uses for the analogous generation poll, which catches the error, increments the attempt counter, sleeps, and only fails once attempts are exhausted:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/f04a8f9953db5d19fb2ea3cbda28a13010fd7679/griptape_nodes_library/proxy/griptape_proxy_node.py#L475-L492

**Impact:** a transient error on any single poll iteration fails an otherwise-successful generation, and the eventual-consistency 404 window makes this most likely on the very first poll, before the asset id is readable.

**Suggested fix:** on exception, log, `await asyncio.sleep(ASSET_POLL_INTERVAL)`, and continue the loop until attempts are exhausted (mirroring `_poll_generation_status`), rather than raising on the first error.

griptape_nodes_library/video/seedance_2_0_video_generation.py
```diff
@@ -422,12 +452,23 @@ def _get_api_model_id(self) -> str:
...
         try:
             await super()._process_generation()
         finally:
             self._public_reference_video_parameter_1.delete_uploaded_artifact()
             self._public_reference_video_parameter_2.delete_uploaded_artifact()
             self._public_reference_video_parameter_3.delete_uploaded_artifact()
+            # The backend deletes the provider asset on terminal generation state, but the
+            # transient GTC static-storage upload made to feed CreateProviderAsset is ours
+            # to clean up, along with the scratch parameter created to perform the upload (its
+            # name is unique per call, so leaving it would accumulate parameters on the node).
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding against the current head of the branch.

**Finding: a registered provider asset can be orphaned when a later build step fails.**

Provider assets are registered eagerly while building the payload, one per reference, before any generation is submitted:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/f04a8f9953db5d19fb2ea3cbda28a13010fd7679/griptape_nodes_library/video/seedance_2_0_video_generation.py#L739-L742

If the first reference registers successfully (its asset reaches ACTIVE) and a later reference's registration, or any other step in `_build_payload`, raises, the payload build aborts and no generation is ever submitted. The cleanup here deletes only the transient GTC static-storage upload; it relies on the backend deleting the provider asset "on terminal generation state", but in this path no generation is ever created, so no terminal state occurs.

**Impact:** the already-registered provider asset(s) leak on the provider side whenever payload construction fails after at least one successful registration.

**Suggested fix:** track successfully-registered provider asset ids and explicitly delete them in this `finally` block when no generation was submitted, or confirm the backend expires unreferenced assets on its own. Flagging as a question, since this depends on backend retention semantics I cannot verify from here.

griptape_nodes_library/video/seedance_human_reference_asset.py
```diff
@@ -0,0 +1,241 @@
...
+
+        self._update_media_visibility()
+
+        # Org-gated feature: probe access at construction so the error surfaces immediately in
+        # the editor as a badge. The probe is best-effort (suppressed, short timeout, fail-closed)
+        # so it never raises out of construction, though it can block for up to the probe timeout.
+        # It is re-probed at run time (validate_before_node_run) because entitlement can change
+        # after the graph is loaded.
+        self._access: ProviderAssetAccess | None = None
+        self._refresh_access()
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding against the current head of the branch.

**Finding: node construction performs blocking, synchronous network I/O.**

`__init__` calls `_refresh_access()`, which runs a synchronous `httpx.get` with a 10s timeout:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/f04a8f9953db5d19fb2ea3cbda28a13010fd7679/griptape_nodes_library/proxy/provider_asset_access.py#L134

Node construction is expected to be cheap and side-effect-free; nodes are instantiated on graph load, add, and copy/paste. A synchronous request that can block for up to `_ACCESS_CHECK_TIMEOUT` (10s) per instance can stall those operations, and if construction runs on the engine's asyncio event loop, a sync `httpx.get` blocks the entire loop and freezes the UI. The same probe also reruns on every `validate_before_node_run`.

Constructing the node was confirmed to fire one synchronous `httpx.get` during `__init__`. The code comment acknowledges it "can block for up to the probe timeout," so the tradeoff is known, but it remains a real responsiveness risk that compounds with multiple nodes or a slow/unreachable endpoint.

**Impact:** graph load, add, and paste can hang for up to 10s per node, and on the event loop the whole UI freezes, whenever the provider-asset endpoint is slow or unreachable.

**Suggested fix:** move the probe off the construction path (lazy/cached on first run, or run it asynchronously and update the badge when it resolves), and/or use a much shorter timeout for the construction-time probe so it can never noticeably block.

griptape_nodes_library/video/seedance_2_0_video_generation.py
zachgiordano: Fixed. `_poll_provider_asset` now logs and retries transient errors (`await asyncio.sleep(ASSET_POLL_INTERVAL)` then `continue`) until attempts are exhausted, mirroring `_poll_generation_status`. Terminal statuses (`FAILED`/`DELETED`) and an `ACTIVE`-without-`asset_id` response still fail immediately, since those are not transient. This also covers the eventual-consistency 404 right after the id is minted.

griptape_nodes_library/video/seedance_2_0_video_generation.py
zachgiordano: Confirmed against the backend — this is handled server-side by design, so no client change is needed (and none is possible: there is no client-facing DELETE endpoint, only `POST /v2/assets` and `GET /v2/assets/<id>`).

The control plane runs a periodic `sweep_orphaned_provider_assets` task that deletes any `ProviderAsset` with `generation__isnull=True` (never linked to a generation) older than `PROVIDER_ASSET_ORPHAN_TTL_HOURS` (default 24h) — explicitly for the "client crashed/aborted between creating the asset and the generation" case. An asset only gets linked to a generation (and thus cleaned up on terminal state) via `link_provider_assets_to_generation` at submit time; anything we register but never submit falls to the orphan sweeper.

So the worst case is an unreferenced asset living up to the TTL before reclamation, which bounds storage on the shared provider account. I updated the cleanup comment to say the backend reclaims orphaned (unsubmitted) assets, rather than implying only the terminal-state path cleans them up.

griptape_nodes_library/video/seedance_human_reference_asset.py
zachgiordano: Fixed by moving the probe off the construction path. `__init__` no longer probes at all — it just initializes the cache. The probe now runs lazily on the first `after_value_set` (cached thereafter) so the editor still gets an access badge as soon as the user interacts with the node, and authoritatively in `validate_before_node_run`. Graph load / add / paste do zero network I/O now. Added unit tests asserting: 0 probes during construction, exactly 1 on first interaction, 0 on subsequent interactions (cached), and the run-time block-on-DENIED behavior.

collindutter (approved):

PR #362, opened 2026-06-26
feat: refreshable variable dropdowns + auto-name for CreateVariable

shhlife:
Closes #206

## Summary

- **GetVariable, SetVariable, HasVariable** — `variable_name` is now a dropdown populated via `ListVariablesRequest` at node creation, with a refresh button (↺) to re-query at any time. The list respects the current scope setting and pre-populates immediately when a node is dropped onto the canvas.

- **CreateVariable** — new `auto_name` boolean that automatically derives the variable name from whatever is connected to `value`:
  - Artifacts (any `*Artifact` type): tries `.name` first, then `.value` for URL artifacts — extracts basename and strips VFX version suffixes (`_v001`, `_001`, etc.)
  - Strings that look like file paths: basename without extension, version suffix stripped
  - Anything else: falls back to source node name (engine `_N` suffix stripped)
  - `auto_name_case` dropdown (hidden until `auto_name` is on): `UPPER CASE`, `UPPER_SNAKE_CASE`, `Title Case`, `snake_case`, `kebab-case`, `PascalCase`, `camelCase`, `as is`

### Create Variable Node

image

### Get Variable Node

image

## Test plan

- [ ] Drop a `GetVariable` / `SetVariable` / `HasVariable` node into a flow that has existing variables — dropdown should pre-populate
- [ ] Click the ↺ button after adding a new variable elsewhere — list should refresh
- [ ] Change scope in Advanced group, click ↺ — list should reflect new scope
- [ ] On `CreateVariable`, enable `auto_name`, connect an `ImageUrlArtifact` from a generation node — variable name should derive from the image filename
- [ ] Try `character_v001.png` and `yellow_house_001.png` — names should strip to `character` / `yellow_house`
- [ ] Switch `auto_name_case` through all options and verify formatting

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/griptape_nodes_library/variables/create_variable.py
+++ b/griptape_nodes_library/variables/create_variable.py
@@ -9,9 +9,17 @@ from griptape_nodes.exe_types.node_types import (
     BaseNode,
     ControlNode,
     NodeDependencies,
+    NodeResolutionState,
     VariableAccess,
     VariableReference,
 )
+from griptape_nodes.retained_mode.events.connection_events import (
+    DeleteConnectionRequest,
+    DeleteConnectionResultSuccess,
+    ListConnectionsForNodeRequest,
+    ListConnectionsForNodeResultSuccess,
+)
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 from griptape_nodes.retained_mode.variable_types import VariableScope
 from griptape_nodes.traits.options import Options
 
@@ -132,14 +140,6 @@ class CreateVariable(ControlNode):
 
     def _delete_incoming_connections_to_parameter(self, parameter_name: str) -> None:
         """Helper to delete all incoming connections to a specific parameter."""
-        from griptape_nodes.retained_mode.events.connection_events import (
-            DeleteConnectionRequest,
-            DeleteConnectionResultSuccess,
-            ListConnectionsForNodeRequest,
-            ListConnectionsForNodeResultSuccess,
-        )
-        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
-
         connections_request = ListConnectionsForNodeRequest(node_name=self.name)
         connections_result = GriptapeNodes.handle_request(connections_request)
 
@@ -162,14 +162,6 @@ class CreateVariable(ControlNode):
 
     def _cleanup_incompatible_value_connections(self) -> None:
         """Remove all connections to/from value parameter that are incompatible with its current type."""
-        from griptape_nodes.retained_mode.events.connection_events import (
-            DeleteConnectionRequest,
-            DeleteConnectionResultSuccess,
-            ListConnectionsForNodeRequest,
-            ListConnectionsForNodeResultSuccess,
-        )
-        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
-
         connections_request = ListConnectionsForNodeRequest(node_name=self.name)
         connections_result = GriptapeNodes.handle_request(connections_request)
 
@@ -251,12 +243,6 @@ class CreateVariable(ControlNode):
 
     def _get_value_source_node_name(self) -> str | None:
         """Return the name of the node wired into the value parameter, or None."""
-        from griptape_nodes.retained_mode.events.connection_events import (
-            ListConnectionsForNodeRequest,
-            ListConnectionsForNodeResultSuccess,
-        )
-        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
-
         try:
             result = GriptapeNodes.handle_request(ListConnectionsForNodeRequest(node_name=self.name))
             if isinstance(result, ListConnectionsForNodeResultSuccess):
@@ -540,8 +526,6 @@ class CreateVariable(ControlNode):
 
     def validate_before_workflow_run(self) -> list[Exception] | None:
         """Variable nodes have side effects and need to execute every workflow run."""
-        from griptape_nodes.exe_types.node_types import NodeResolutionState
-
         self.make_node_unresolved(
             current_states_to_trigger_change_event={NodeResolutionState.RESOLVED, NodeResolutionState.RESOLVING}
         )
@@ -549,8 +533,6 @@ class CreateVariable(ControlNode):
 
     def validate_before_node_run(self) -> list[Exception] | None:
         """Variable nodes have side effects and need to execute every time they run."""
-        from griptape_nodes.exe_types.node_types import NodeResolutionState
-
         self.make_node_unresolved(
             current_states_to_trigger_change_event={NodeResolutionState.RESOLVED, NodeResolutionState.RESOLVING}
         )
--- a/griptape_nodes_library/variables/create_variable.py
+++ b/griptape_nodes_library/variables/create_variable.py
@@ -312,20 +312,24 @@ class CreateVariable(ControlNode):
     @staticmethod
     def _apply_case(words: list[str], case_style: str) -> str:
         """Join tokens using the requested case convention."""
-        if case_style == CaseStyle.UPPER:
-            return " ".join(w.upper() for w in words)
-        if case_style == CaseStyle.UPPER_SNAKE:
-            return "_".join(w.upper() for w in words)
-        if case_style == CaseStyle.TITLE:
-            return " ".join(w.capitalize() for w in words)
-        if case_style == CaseStyle.KEBAB:
-            return "-".join(w.lower() for w in words)
-        if case_style == CaseStyle.PASCAL:
-            return "".join(w.capitalize() for w in words)
-        if case_style == CaseStyle.CAMEL:
-            return words[0].lower() + "".join(w.capitalize() for w in words[1:])
-        # Default: snake_case
-        return "_".join(w.lower() for w in words)
+        match case_style:
+            case CaseStyle.UPPER:
+                return " ".join(w.upper() for w in words)
+            case CaseStyle.UPPER_SNAKE:
+                return "_".join(w.upper() for w in words)
+            case CaseStyle.TITLE:
+                return " ".join(w.capitalize() for w in words)
+            case CaseStyle.SNAKE:
+                return "_".join(w.lower() for w in words)
+            case CaseStyle.KEBAB:
+                return "-".join(w.lower() for w in words)
+            case CaseStyle.PASCAL:
+                return "".join(w.capitalize() for w in words)
+            case CaseStyle.CAMEL:
+                return words[0].lower() + "".join(w.capitalize() for w in words[1:])
+            case _:
+                msg = f"Unknown case style: {case_style!r}"
+                raise ValueError(msg)
 
     @staticmethod
     def _strip_version_suffix(stem: str) -> str:
--- a/griptape_nodes_library/variables/create_variable.py
+++ b/griptape_nodes_library/variables/create_variable.py
@@ -1,7 +1,10 @@
 import logging
+import os
+import re
 from typing import Any
 
 from griptape_nodes.exe_types.core_types import Parameter, ParameterMode, ParameterTypeBuiltin
+from griptape_nodes.traits.options import Options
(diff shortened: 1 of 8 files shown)


griptape_nodes_library/variables/create_variable.py
```diff
@@ -10,9 +12,20 @@
...
+CASE_UPPER = "UPPER CASE"
+CASE_UPPER_SNAKE = "UPPER_SNAKE_CASE"
+CASE_TITLE = "Title Case"
+CASE_SNAKE = "snake_case"
+CASE_KEBAB = "kebab-case"
+CASE_PASCAL = "PascalCase"
+CASE_CAMEL = "camelCase"
+CASE_AS_IS = "as is"
+CASE_OPTIONS = [CASE_UPPER, CASE_UPPER_SNAKE, CASE_TITLE, CASE_SNAKE, CASE_KEBAB, CASE_PASCAL, CASE_CAMEL, CASE_AS_IS]
+
```
SavagePencil: if you declare these as a StrEnum, is there a way to enumerate them automagically?

griptape_nodes_library/variables/create_variable.py
```diff
@@ -192,6 +232,147 @@ def _cleanup_incompatible_value_connections(self) -> None:
...
+                self.set_parameter_value(self.variable_name_param.name, derived)
+
+        super().after_value_set(parameter, value)
+
+    def _get_value_source_node_name(self) -> str | None:
+        """Return the name of the node wired into the value parameter, or None."""
+        from griptape_nodes.retained_mode.events.connection_events import (
+            ListConnectionsForNodeRequest,
+            ListConnectionsForNodeResultSuccess,
+        )
```
SavagePencil: nit: I think these two could be un-lazied?

griptape_nodes_library/variables/create_variable.py
```diff
@@ -192,6 +232,147 @@ def _cleanup_incompatible_value_connections(self) -> None:
...
+            result = GriptapeNodes.handle_request(ListConnectionsForNodeRequest(node_name=self.name))
+            if isinstance(result, ListConnectionsForNodeResultSuccess):
+                for conn in result.incoming_connections:
+                    if conn.target_parameter_name == self.value_param.name:
+                        return conn.source_node_name
+        except Exception:
+            pass
+        return None
+
+    def _derive_variable_name(self, value: Any, source_node_name: str | None) -> str | None:
```
SavagePencil, line 256: wondering if you could do macros to do the extraction, and possibly also the styling of them?

griptape_nodes_library/variables/create_variable.py
```diff
@@ -192,6 +232,147 @@ def _cleanup_incompatible_value_connections(self) -> None:
...
+        if case_style == CASE_TITLE:
+            return " ".join(w.capitalize() for w in words)
+        if case_style == CASE_KEBAB:
+            return "-".join(w.lower() for w in words)
+        if case_style == CASE_PASCAL:
+            return "".join(w.capitalize() for w in words)
+        if case_style == CASE_CAMEL:
+            return words[0].lower() + "".join(w.capitalize() for w in words[1:])
+        # Default: snake_case
+        return "_".join(w.lower() for w in words)
```
SavagePencil: match/case this so you can catch and surface as an error an invalid type

griptape_nodes_library/variables/create_variable.py
```diff
@@ -10,9 +12,20 @@
     VariableReference,
 )
 from griptape_nodes.retained_mode.variable_types import VariableScope
+from griptape_nodes.traits.options import Options
 
 logger = logging.getLogger("griptape_nodes")
 
+CASE_UPPER = "UPPER CASE"
```
SavagePencil: coudl we add these to the list of macro FormatSpecs and let the macro system do it all for us?

griptape_nodes_library/variables/create_variable.py
```diff
@@ -192,6 +232,147 @@ def _cleanup_incompatible_value_connections(self) -> None:
...
+            ".avi",
+            ".mp3",
+            ".wav",
+            ".ogg",
+            ".json",
+            ".txt",
+            ".csv",
+            ".pdf",
+        }
+        _, ext = os.path.splitext(s)
```
SavagePencil: could we query the ArtifactManager for all of the extensions it supports? Could be a TODO

griptape_nodes_library/variables/create_variable.py
```diff
@@ -192,6 +232,147 @@ def _cleanup_incompatible_value_connections(self) -> None:
...
+    @staticmethod
+    def _strip_version_suffix(stem: str) -> str:
+        """Strip VFX-convention version/frame suffixes from a filename stem.
+
+        character_v001 → character, yellow_house_v001 → yellow_house,
+        yellow_house_001 → yellow_house. Single trailing digits (texture2d) are left alone.
+        """
+        stem = re.sub(r"[_\-]v\d+$", "", stem, flags=re.IGNORECASE)
+        stem = re.sub(r"[_\-]\d{2,}$", "", stem)
+        return stem
```
SavagePencil, line 343: see if macro system can do this for us?

griptape_nodes_library/variables/create_variable.py
shhlife: adding a rule in CLAUDE.md about it, too

griptape_nodes_library/variables/create_variable.py
shhlife: adding a rule in CLAUDE.md about it, too

griptape_nodes_library/variables/create_variable.py
```diff
@@ -192,6 +226,141 @@ def _cleanup_incompatible_value_connections(self) -> None:
...
+
+        Priority:
+        1. Artifact with a filename (ImageArtifact, AudioArtifact, VideoArtifact, BlobArtifact)
+        2. String that looks like a file path → basename without extension
+        3. Source node name (engine _N suffix stripped)
+        """
+        # Any artifact type: try .name first (filename-bearing artifacts like ImageArtifact),
+        # then .value (URL-bearing artifacts like ImageUrlArtifact). Guards on class name so we
+        # don't accidentally match plain dicts or strings that happen to have a .name attr.
+        if "Artifact" in type(value).__name__:
```
SavagePencil, line 280: we might be able to use the ArtifactManager to see if it handles these? It may not. If so ignore.

griptape_nodes_library/variables/create_variable.py
shhlife: @SavagePencil here's what claude said:

The FormatSpec system in `griptape_nodes.common.macro_parser.formats` has these transforms: `LowerCaseFormat`, `UpperCaseFormat`, `SlugFormat`, `SeparatorFormat`, N`umericPaddingFormat`, `DateFormat`. The `FORMAT_REGISTRY` exposes `lower`, `upper`, and `slug`.

For name extraction: Not applicable — FormatSpecs are string transforms, not extractors. They don't handle artifact duck-typing, file path stemming, or version suffix stripping.

For case styling: Partial overlap only. Of our 8 CaseStyle options:

UPPER CASE → could use UpperCaseFormat on space-joined words
kebab-case → SlugFormat is close but also strips non-alphanumeric chars (we just hyphen-join)
snake_case, UPPER_SNAKE_CASE, Title Case, PascalCase, camelCase, as is → not in the registry at all
To use FormatSpecs for case styling we'd need to add SnakeCaseFormat, PascalCaseFormat, CamelCaseFormat, TitleCaseFormat, UpperSnakeCaseFormat to the upstream macro system — a significant change to a different package, for transforms that are only useful here and not anywhere else in the macro path template system.

The current `_apply_case()` + `_tokenize()` in `create_variable.py` is already the right implementation for this use case. FormatSpecs are designed for path template variable substitution (e.g., `{filename:lower}, {version:03}`), not for word-list case transformation.

Short answer: the macro FormatSpec system doesn't cover the case styles we need, and extending it just for this would add complexity without any benefit to the macro system itself. The custom _apply_case() is the right call.

if we really want this, we could create another followup task?

griptape_nodes_library/variables/create_variable.py
shhlife: @SavagePencil  can this be a TODO for future us to modify the engine and then update after?

griptape_nodes_library/variables/create_variable.py
SavagePencil: No no no, Claude! You got so confused!

I meant can we use the macro system to do the extraction of the tasty bits we care about instead of a custom regex (now you have two problems...). Then when we have the tasty bits, we could feed those variables into ANOTHER macro that potentially uses the FormatSpecs to stylize them?

Alternatively, if you agree with this approach:
1. Do extraction from matching customer input to a macro to get the tasty bits
2. Keep custom rendering, but add a TODO to use a macro to render it later once the missing FormatSpecs get added.

griptape_nodes_library/variables/create_variable.py
SavagePencil: yup

SavagePencil: @shhlife it was more a question of "would doing that even be smart", not an ask for this. But if it can and should, then by all means add a TODO.

griptape_nodes_library/variables/create_variable.py
shhlife: checking..

shhlife: actually.. we don't really _need_ to get valid file extensions.. this just just a handy quick thing to trim the extension if it exists.. we'll just use a simple regex.

SavagePencil: you mean macro match, right :)

(if it fits the bill)

griptape_nodes_library/variables/create_variable.py
shhlife: @griptapeops can you create a task to update the FormatSpec to handle more types? pls and thank you kindly

griptape_nodes_library/variables/create_variable.py
shhlife: claude: That's a sharp observation — and you're right that when the macro system extracts variables from a path (e.g., matching character_v001 against {name}_v{version:03}), it effectively strips the version suffix as a side effect of parsing. So internally, yes, that logic exists.

But the problem is that _strip_version_suffix handles patterns the macro system can't without upfront template knowledge:

_v001, _v0001, _v1 — any digit count (our regex [_\-]v\d+$ handles all of these; NumericPaddingFormat only handles a specific fixed width like :03)
_001 — no v prefix at all
To replicate _strip_version_suffix with the macro system you'd have to try multiple templates in sequence ({name}_v{version:03}, {name}_v{version:04}, {name}_{version:03}, …) and use the first match. That's more code and more brittle than the two regexes.

If the project ever standardizes on a single version format (e.g., always _v####), the macro system would be the right tool and _strip_version_suffix could be retired. For now, the regexes handle the messy reality of mixed VFX conventions that the macro system wasn't designed for.

griptape_nodes_library/variables/create_variable.py
shhlife: claude says:

No — ArtifactManager is a preview/format system. Its public surface is:

sniff_extension(data: bytes) — magic-byte detection from raw file bytes
prepare_content_for_write(data, file_name) — format-specific pre-write processing
get_artifact_metadata(source_path) — extracts things like video FPS/dimensions from a file on disk

None of that touches semantic name extraction from a path string. The current code — checking artifact.name, then falling back to parsing artifact.value as a path — is the right approach and there's nothing in ArtifactManager that would help.

griptape_nodes_library/variables/create_variable.py
SavagePencil: You right, my dude

SavagePencil (approved):

PR #374, opened 2026-06-30
feature: Add Random Number node (#373)

shhlife:
## Summary

- Adds a new `RandomNumber` node to `griptape_nodes_library/number/`
- Registered in `griptape_nodes_library.json` under the `number` category, `create` group

## Details

The node exposes:
- **mode** — dropdown to select `integer` or `float` output
- **minimum / maximum** — range inputs (default 0–100), accept `int` or `float`
- **randomize_seed** — when enabled, generates a fresh seed each run
- **seed** — fixed seed (default 42) for reproducible results; field becomes read-only when `randomize_seed` is on
- **result** — the generated random number

Uses `SeedParameter` for the seed/randomize pattern (same as image/video generation nodes) and `async def aprocess` so `preprocess()` is called at execution time to apply the randomized seed before generating the number.

Closes #373

## Test plan

- [ ] Add node to a workflow; run it — result appears in the `result` output
- [ ] Set `mode = integer`, verify result is a whole number in [minimum, maximum]
- [ ] Set `mode = float`, verify result is a decimal in [minimum, maximum)
- [ ] Set a fixed seed, run multiple times — same result each time
- [ ] Enable `randomize_seed`, run multiple times — different results each run; seed field becomes read-only

### DIFF
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -4989,6 +4989,24 @@
       }
     },
     {
+      "class_name": "RandomNumber",
+      "file_path": "griptape_nodes_library/number/random_number.py",
+      "metadata": {
+        "category": "number",
+        "description": "Generate a random integer or float within a given range, with an optional seed for reproducibility.",
+        "display_name": "Random Number",
+        "icon": "dices",
+        "group": "create",
+        "tags": [
+          "number",
+          "random",
+          "integer",
+          "float",
+          "generate"
+        ]
+      }
+    },
+    {
       "class_name": "Ruleset",
       "file_path": "griptape_nodes_library/rules/create_ruleset.py",
       "metadata": {
--- a/griptape_nodes_library/number/random_number.py
+++ b/griptape_nodes_library/number/random_number.py
@@ -1,0 +1,83 @@
+import random
+from enum import StrEnum
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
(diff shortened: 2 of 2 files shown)


griptape_nodes_library.json
```diff
@@ -29,7 +29,7 @@
   "metadata": {
     "author": "Griptape, Inc",
     "description": "Default nodes for Griptape Nodes",
-    "library_version": "0.80.1",
+    "library_version": "0.80.2",
```
collindutter: Don't bump in the PR, will bump on release.

collindutter (changes requested):

griptape_nodes_library.json
shhlife: @collindutter it keeps giving me version-check errors if we don't bump it

image

collindutter: Ah that came from https://github.com/griptape-ai/griptape-nodes-library-standard/pull/364 which we have a thread on. I reverted that, pull in the latest to your branch and it should be good.

shhlife: I did.. this was freshly pulled

collindutter: I reverted seconds ago, are you sure you pulled that commit in from main? That script doesn't exist on main.

shhlife: ah, okay - will try again

shhlife: @collindutter okay - that took some surgery. :) thanks for the help claude!

collindutter (approved):

PR #377, opened 2026-06-30
fix: Agent breaking because of ProviderConfig

shhlife:
## Summary

- Fixes `NameError: name 'ProviderConfig' is not defined` on the Agent node when provider events are not available in the running engine version
- Quotes the two method annotations that reference `ProviderConfig` (`_fetch_providers` return type and `_resolve_provider_api_key` parameter type) so they are evaluated lazily rather than at class-definition time

## Root cause

`ProviderConfig` is imported inside a `try/except ImportError` guard, so it may not be bound if the host engine doesn't expose `agent_events`. Python evaluates method annotations eagerly when the class body is executed (module import time), so the unbound name caused a `NameError` before any runtime guard could fire. Converting the annotations to string forward references (`"ProviderConfig"`) defers evaluation until the annotations are actually inspected.

Fixes #376

### DIFF
--- a/griptape_nodes_library/agents/agent.py
+++ b/griptape_nodes_library/agents/agent.py
@@ -78,6 +78,7 @@ try:
         ListProviderModelsResultSuccess,  # pyright: ignore[reportAttributeAccessIssue]
         ProviderConfig,  # pyright: ignore[reportAttributeAccessIssue]
     )
+
     _AGENT_PROVIDERS_AVAILABLE = True
     _GRIPTAPE_CLOUD_PROVIDER = ProviderConfig(name="griptape_cloud", type="griptape_cloud", model="")
 except ImportError:
--- a/griptape_nodes_library/agents/agent.py
+++ b/griptape_nodes_library/agents/agent.py
@@ -7,7 +7,7 @@ for tools, rulesets, prompts, and streams output back to the user interface.
 """
 
 import json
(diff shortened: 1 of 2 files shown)


griptape_nodes_library/agents/agent.py
```diff
@@ -341,7 +341,7 @@ def _build_tool_exchange(self, subtask_events: list) -> str:
 
     # --- Provider / Model Methods ---
 
-    def _fetch_providers(self) -> list[ProviderConfig]:  # pyright: ignore[reportPossiblyUnbound,reportPossiblyUnboundVariable]
+    def _fetch_providers(self) -> "list[ProviderConfig]":  # pyright: ignore[reportPossiblyUnbound,reportPossiblyUnboundVariable]
```
collindutter: Is this what `# pyright: ignore[reportPossiblyUnbound,reportPossiblyUnboundVariable]` was warning us about? Can we resolve?

uv.lock
collindutter, line 1: Should not be checked in

collindutter (changes requested):

griptape_nodes_library/agents/agent.py
shhlife: @collindutter can you check it out again? thanks!

collindutter (approved):

PR #382, opened 2026-06-30
Tripo Multiview Node

cjkindel:
~Unblocked by https://github.com/griptape-ai/griptape-cloud/pull/2065~

Closes https://github.com/griptape-ai/internal/issues/80

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -748,7 +748,7 @@ wheels = [
 
 [[package]]
 name = "griptape-nodes-library-standard"
-version = "0.67.0"
+version = "0.80.1"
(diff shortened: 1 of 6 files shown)


collindutter (approved):

uv.lock
collindutter, line 1: Intended?

cjkindel: No... This should be 81.0 anyways, right?

collindutter: It should be. But we should fix in the pyproject too. Fine with tacking that onto this PR.

collindutter (approved):

PR #384, opened 2026-07-01
adds html

shhlife:
adds html as a supported text extension
fixes: #378


griptape_nodes_library/utils/file_utils.py
```diff
@@ -10,6 +10,7 @@
     ".json",
     ".yaml",
     ".yml",
+    ".html",
```
cjkindel, line 14: And .htm?

cjkindel (approved):

PR #397, opened 2026-07-04
feat: add situation dropdown to all save nodes

### LINKED ISSUE #330: Save to Project: let users pick the situation on the node instead of opening the config
## Problem

The `Save to Project` node (`SaveToProject`) always writes using a single hardcoded situation — `save_node_output`. If you want files from this node to land via a different situation/path template, there's no way to pick it on the node itself. You have to click the cog to spin up a `FileOutputSettings` node and wire it into the `output_file` parameter just to change the situation.

That's a lot of steps for what is conceptually a single dropdown choice, and it clutters the graph with an extra node. For a save node where choosing *where things go* is the whole point, selecting the situation should be fast and inline.

## Origin

Reported by a team member working in the Nodes editor. The friction is the config detour: setting the save situation requires opening the configuration (creating + connecting a `FileOutputSettings` node) rather than choosing it directly on the `Save to Project` node.

## Impact

Adds repeated friction to a common operation. Every time a user wants a non-default situation they create and connect an extra node, which is time-consuming and makes simple graphs busier than they need to be.

## Possible direction

One option: add a `situation` selector (e.g. a dropdown of available situations) directly on the `Save to Project` node, defaulting to the current `save_node_output`. The existing `FileOutputSettings` connection path would still work and take precedence when connected, for cases that need the fuller config. Open questions worth pinning down:

- Where the list of selectable situations comes from (project-defined situations vs. a fixed set).
- Precedence when both an inline situation and a connected `FileOutputSettings` node are present (proposal: connection wins, matching today's behavior).
- Whether this should apply only to `Save to Project` or to every node built on `ProjectFileParameter` (e.g. `save_image`, `save_audio`, `save_video`, `save_text`, `save_dict`) so the affordance is consistent.

shhlife:
Closes #330

# 🔥 Waiting for https://github.com/griptape-ai/griptape-nodes-engine/pull/5047

image

## Summary

- Adds an inline **Situation** dropdown to all five save nodes (`SaveImage`, `SaveAudio`, `SaveVideo`, `SaveText`, `SaveToProject`) — no more detour through `FileOutputSettings` just to change where files land
- Situation dropdown also added to `FileOutputSettings` itself, with descriptions shown as subtitles (engine 0.91.0+, gracefully degraded on 0.90.0 via `getattr`)
- Refresh button on save node dropdowns lets users pick up newly-added situations without reloading
- Shared logic DRY'd into `griptape_nodes_library/utils/situation_utils.py` — `add_situation_parameter()`, `fetch_situations()`, `build_situation_data()`, `update_file_param_situation()`
- Uses `GetAllSituationsForProjectRequest` (init-safe) to avoid the `reentrant-bus-in-init` strict-mode rule
- Adds `.claude/skills/dropdown-row-subtitles` skill documenting the `update_ui_options` subtitle + icon pattern

## Engine compatibility

`fetch_situations()` is written to work across three engine generations:

| Engine | `descriptions` field | `fields` param | Behavior |
|--------|---------------------|----------------|----------|
| 0.90.x | ✗ | ✗ | Names only (subtitles blank) |
| 0.91.x | ✓ (via `getattr`) | ✗ | Names + descriptions |
| 0.93.x+ | ✓ | ✓ ([#5047](https://github.com/griptape-ai/griptape-nodes-engine/pull/5047)) | Names + descriptions, filtered fetch |

The `fields=["situations", "descriptions"]` call falls back gracefully to an unfiltered request on older engines via `TypeError`.

## Test plan

- [ ] Add a Save Image / Save Audio / Save Video / Save Text / Save to Project node — situation dropdown appears pre-populated with the project's situations
- [ ] Change situation on a save node and run — output lands in the correct situation directory
- [ ] Click the refresh button after adding a new situation to the project — dropdown updates without reloading
- [ ] Open a FileOutputSettings node — situation dropdown shows names (and descriptions as subtitles on engine 0.91.0+)
- [ ] Workflow round-trip: save and reload — chosen situation is preserved

### DIFF
--- a/griptape_nodes_library/utils/situation_utils.py
+++ b/griptape_nodes_library/utils/situation_utils.py
@@ -1,0 +1,45 @@
+"""Shared helpers for situation dropdown parameters in save nodes."""
+
+import logging
+
+from griptape_nodes.exe_types.param_components.project_file_parameter import ProjectFileParameter
+from griptape_nodes.retained_mode.events.project_events import (
+    GetCurrentProjectRequest,
+    GetCurrentProjectResultSuccess,
+    GetProjectTemplateRequest,
+    GetProjectTemplateResultSuccess,
+)
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+logger = logging.getLogger("griptape_nodes")
+
+DEFAULT_SITUATION = ProjectFileParameter.DEFAULT_SITUATION
+
+
+def fetch_situations_with_descriptions() -> tuple[list[str], dict[str, str]]:
+    """Return (sorted_names, {name: description}) from the current project.
+
+    Falls back to a single-entry list containing the default situation when the
+    project template is not yet loaded.
+    """
+    current = GriptapeNodes.handle_request(GetCurrentProjectRequest())
+    if not isinstance(current, GetCurrentProjectResultSuccess):
+        logger.warning("Could not fetch current project; using default situation")
+        return [DEFAULT_SITUATION], {}
+
+    template_result = GriptapeNodes.handle_request(
+        GetProjectTemplateRequest(project_id=current.project_info.project_id)
+    )
+    if not isinstance(template_result, GetProjectTemplateResultSuccess):
+        logger.warning("Could not fetch project template; using default situation")
+        return [DEFAULT_SITUATION], {}
+
+    situations = template_result.template.situations
+    names = sorted(situations.keys())
+    descriptions = {name: (sit.description or "") for name, sit in situations.items()}
+    return names, descriptions
+
+
+def build_situation_data(names: list[str], descriptions: dict[str, str]) -> list[dict]:
+    """Build the ui_options data list for dropdown_row_subtitles display."""
+    return [{"name": n, "subtitle": descriptions.get(n, "")} for n in names]
--- a/griptape_nodes_library/utils/situation_utils.py
+++ b/griptape_nodes_library/utils/situation_utils.py
@@ -1,9 +1,9 @@
 """Shared helpers for situation dropdown parameters in save nodes."""
 
 import logging
-from typing import Any
 
-from griptape_nodes.exe_types.core_types import NodeMessageResult, ParameterMode
+from griptape_nodes.exe_types.core_types import NodeMessageResult, Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import BaseNode
 from griptape_nodes.exe_types.param_components.project_file_parameter import ProjectFileParameter
 from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
 from griptape_nodes.retained_mode.events.project_events import (
@@ -50,7 +50,21 @@ def build_situation_data(names: list[str], descriptions: dict[str, str]) -> list
     return [{"name": n, "subtitle": descriptions.get(n, "")} for n in names]
 
 
-def add_situation_parameter(node: Any, file_param: ProjectFileParameter) -> None:
+def update_file_param_situation(
+    file_param: ProjectFileParameter, parameter: Parameter, value: object, **kwargs: object
+) -> None:
+    """Sync situation name to file_param when the situation parameter changes.
+
+    Centralises the single write to the private ``_situation_name`` attribute so
+    all five save nodes share one call site instead of five.  The ``initial_setup``
+    guard mirrors ``FileOutputSettings.after_value_set`` — skip internal framework
+    setup calls because ``add_situation_parameter`` already seeds the correct default.
+    """
+    if parameter.name == "situation" and not kwargs.get("initial_setup"):
+        file_param._situation_name = str(value)
+
+
+def add_situation_parameter(node: BaseNode, file_param: ProjectFileParameter) -> None:
     """Add a situation dropdown and refresh button to a save node.
 
     ``file_param`` must be created (but NOT yet added via ``add_parameter()``)
@@ -116,6 +130,8 @@ def add_situation_parameter(node: Any, file_param: ProjectFileParameter) -> None
         settable=True,
     )
     node.add_parameter(situation_param)
+    # update_ui_options must follow add_parameter — subtitle data is separate from
+    # the trait and must be pushed to the UI layer after the param is registered.
     situation_param.update_ui_options({
         "data": build_situation_data(names, descriptions),
         "dropdown_row_subtitles": True,
--- a/griptape_nodes_library/utils/situation_utils.py
+++ b/griptape_nodes_library/utils/situation_utils.py
@@ -7,6 +7,8 @@ from griptape_nodes.exe_types.node_types import BaseNode
 from griptape_nodes.exe_types.param_components.project_file_parameter import ProjectFileParameter
 from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
 from griptape_nodes.retained_mode.events.project_events import (
+    GetAllSituationsForProjectRequest,
+    GetAllSituationsForProjectResultSuccess,
     GetCurrentProjectRequest,
     GetCurrentProjectResultSuccess,
     GetProjectTemplateRequest,
@@ -19,6 +21,21 @@ from griptape_nodes.traits.options import Options
 logger = logging.getLogger("griptape_nodes")
 
 DEFAULT_SITUATION = ProjectFileParameter.DEFAULT_SITUATION
+
+
+def fetch_situation_names() -> list[str]:
+    """Return sorted situation names from the current project.
+
+    Uses GetAllSituationsForProjectRequest, which is safe to call from __init__
+    (does not trigger the reentrant-bus-in-init strict-mode rule). Falls back to
+    [DEFAULT_SITUATION] when the project template is not yet loaded.
+    """
+    result = GriptapeNodes.handle_request(GetAllSituationsForProjectRequest())
+    if not isinstance(result, GetAllSituationsForProjectResultSuccess):
+        logger.warning("Could not fetch situation names; using default situation")
+        return [DEFAULT_SITUATION]
+    names = sorted(result.situations.keys())
+    return names if names else [DEFAULT_SITUATION]
 
 
 def fetch_situations_with_descriptions() -> tuple[list[str], dict[str, str]]:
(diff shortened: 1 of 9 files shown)


griptape_nodes_library/utils/situation_utils.py
```diff
@@ -0,0 +1,135 @@
...
+) -> None:
+    """Sync situation name to file_param when the situation parameter changes.
+
+    Centralises the single write to the private ``_situation_name`` attribute so
+    all five save nodes share one call site instead of five.  The ``initial_setup``
+    guard mirrors ``FileOutputSettings.after_value_set`` — skip internal framework
+    setup calls because ``add_situation_parameter`` already seeds the correct default.
+    """
+    if parameter.name == "situation" and not kwargs.get("initial_setup"):
+        file_param._situation_name = str(value)
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding.

**Finding: saved `situation` selection is silently ignored after workflow reload.**

`_situation_name` is synced from the `situation` parameter only here, via `update_file_param_situation`, which each node calls from `after_value_set`. On workflow load the engine restores parameter values with `initial_setup=True` and, in that path, skips `after_value_set` entirely. Because `situation` is a PROPERTY-only parameter (no upstream connection), it is also never re-hydrated at run time. So after a save/reload `_situation_name` stays at the constructor default (`DEFAULT_SITUATION` or `names[0]`) while the dropdown still displays the user's saved choice.

Live edits within the same session work (they hit `after_value_set` with `initial_setup=False`), which makes this easy to miss during manual testing.

Note that `FileOutputSettings` sidesteps this whole class of bug by reading the value at resolve time instead of caching it via a hook:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/38a9397c935aaafc37e1cdfc95a598b62ff1c09c/griptape_nodes_library/filesystem/file_output_settings.py#L242-L244

**Impact:** `build_file()` writes using the *default* situation while the UI shows the user's selection, so files land in the wrong directory / with the wrong naming after any reload. This is a silent correctness bug on the exact feature this PR adds.

**Suggested fix:** resolve the situation at build/process time rather than caching it through a hook, e.g. set `self._output_file._situation_name = self.get_parameter_value("situation")` at the top of each node's `process`/`aprocess`. The more robust fix is to move the whole situation concern into `ProjectFileParameter` (a situation getter it resolves fresh in `build_file()`), which also eliminates the per-node hand-threading of three coordinated steps across the five save nodes.

griptape_nodes_library/utils/situation_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    def _on_refresh(_button: Button, button_details: ButtonDetailsMessagePayload) -> NodeMessageResult:
+        refreshed_names, refreshed_descriptions = fetch_situations()
+
+        if not refreshed_names:
+            return NodeMessageResult(
+                success=False,
+                details="No situations available — project template could not be loaded",
+                response=button_details,
+                altered_workflow_state=False,
+            )
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding.

**Finding: unreachable failure branch in `_on_refresh`.**

`fetch_situations` never returns an empty list: on both the non-success path and the empty-keys path it falls back to `[DEFAULT_SITUATION]`. So `refreshed_names` is always non-empty and this `if not refreshed_names:` block can never execute.

**Impact:** the "No situations available" error is dead code, and a genuinely failed template load is masked behind the default situation rather than being surfaced to the user.

**Suggested fix:** either delete this branch, or have `fetch_situations` signal a real failure (a sentinel or a success flag) instead of always substituting the default, so this branch can actually distinguish "template failed to load" from "template has no situations."

griptape_nodes_library/utils/situation_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    """Build the ui_options data list for dropdown_row_subtitles display."""
+    return [{"name": n, "subtitle": descriptions.get(n, "")} for n in names]
+
+
+def update_file_param_situation(
+    file_param: ProjectFileParameter, parameter: Parameter, value: object, **kwargs: object
+) -> None:
+    """Sync situation name to file_param when the situation parameter changes.
+
+    Centralises the single write to the private ``_situation_name`` attribute so
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding.

**Finding: docstring overstates the abstraction, and the module pokes an engine-private attribute from three sites.**

This docstring says it "centralises the *single* write ... so all five save nodes share *one* call site," but `_situation_name` is actually written from three places in this module:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/38a9397c935aaafc37e1cdfc95a598b62ff1c09c/griptape_nodes_library/utils/situation_utils.py#L60-L61

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/38a9397c935aaafc37e1cdfc95a598b62ff1c09c/griptape_nodes_library/utils/situation_utils.py#L101-L101

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/38a9397c935aaafc37e1cdfc95a598b62ff1c09c/griptape_nodes_library/utils/situation_utils.py#L135-L135

Two further points on the `initial_setup` guard: against the pinned engine it is inert (the engine never passes `initial_setup` to `after_value_set`; it simply skips the call during initial setup), so `kwargs` is always empty and the guard never fires. The docstring's claim that this "mirrors `FileOutputSettings.after_value_set`" is misleading because that guard is equally inert. There is also latent fragility: if `kwargs` ever did carry `initial_setup`, the `super().after_value_set(parameter, value, **kwargs)` call in each node would raise `TypeError`, since the base signature is only `(parameter, value)`.

**Impact:** the library is coupled to a private engine attribute (`_situation_name`) that can change without warning, and the docstring misrepresents both the number of write sites and what the guard does.

**Suggested fix:** add a public setter/getter on `ProjectFileParameter` (engine side) and route all three writes through it; correct the docstring. Combined with the build-time-resolution fix, the cleanest shape is for `ProjectFileParameter` to own a situation getter so `build_file()` resolves it fresh and no external attribute-poking is needed.

griptape_nodes_library/utils/situation_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    all five save nodes share one call site instead of five.  The ``initial_setup``
+    guard mirrors ``FileOutputSettings.after_value_set`` — skip internal framework
+    setup calls because ``add_situation_parameter`` already seeds the correct default.
+    """
+    if parameter.name == "situation" and not kwargs.get("initial_setup"):
+        file_param._situation_name = str(value)
+
+
+def add_situation_parameter(node: BaseNode, file_param: ProjectFileParameter) -> None:
```
collindutter, line 74: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding.

**Finding: two competing situation controls with silent precedence.**

This dropdown is added to every save node, but when an upstream `FileOutputSettings` is connected to `output_file`, `build_file()` uses the provider's situation and ignores the save node's `situation` dropdown entirely. The dropdown nonetheless stays visible and editable, so a user can change it and observe no effect.

**Impact:** confusing UX. The visible control silently does nothing whenever the incoming connection exists, with no indication of which control wins.

**Suggested fix:** hide or disable the `situation` parameter when an incoming connection to `output_file` is present. `ProjectFileParameter` already hooks `on_incoming_connection_removed`, so wiring the connection-added side to toggle the dropdown's visibility is a natural companion.

griptape_nodes_library/utils/situation_utils.py
collindutter: I'd prefer the engine side change.

collindutter (changes requested):

griptape_nodes_library/utils/situation_utils.py
shhlife: @collindutter  would you prefer to hold this and fix it in engine first, or would you be okay with a library fix and engine followup?

collindutter: Fine with library for now.

shhlife: adding a note & the followup ticket

collindutter (changes requested):

griptape_nodes_library/utils/situation_utils.py
```diff
@@ -0,0 +1,138 @@
...
+DEFAULT_SITUATION = ProjectFileParameter.DEFAULT_SITUATION
+_OUTPUT_FILE_PARAM = "output_file"
+
+
+def on_output_file_connected(node: BaseNode, target_parameter: Parameter) -> None:
+    """Call from after_incoming_connection to hide the situation dropdown."""
+    if target_parameter.name == _OUTPUT_FILE_PARAM:
+        param = node.get_parameter_by_name("situation")
+        if param is not None:
+            param.hide = True
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding.

**Finding: situation dropdown is over-hidden for plain-string connections to `output_file`.**

`on_output_file_connected` hides the `situation` dropdown for **any** incoming connection to `output_file`, keying only on the parameter name. But `ProjectFileParameter.build_file()` only bypasses the situation when the upstream node is a `FileDestinationProvider`; for a plain `str` source it still resolves the path through `self._situation_name`:

```python
if isinstance(source_node, FileDestinationProvider):
    ...
    return file_dest
# plain-string source falls through to:
return ProjectFileDestination.from_situation(filename, self._situation_name, **extra_vars)
```

`output_file` accepts `INPUT` connections from any `str` source, so wiring e.g. a text node's output into `output_file` hides the situation control even though the situation still governs where the file lands.

**Impact:** this is worse than the original "visible but ignored" case. The control now disappears while remaining in effect, so the user can neither see nor change a setting that is still applied to their save path.

**Suggested fix:** gate the hide on the actual source type. Pass `source_node` into the helper and hide only when it is a `FileDestinationProvider` (`@runtime_checkable`, importable from `griptape_nodes.files.file`) — the same predicate `build_file()` uses. Each override already has `source_node` in hand.

collindutter (changes requested):

griptape_nodes_library/utils/situation_utils.py
shhlife: hah! can't believe I missed that

collindutter (approved):

PR #412, opened 2026-07-07
fix(tests): restore flow inputs and metadata preset for integration workflows

collindutter:
The Workflow Tests job on main has been failing 3 tests for repo-side reasons (the other 3 Tripo failures were upstream credit exhaustion, since resolved):

- `test_openai_image_generation_wide` / `test_openai_image_generation_tall` were added in #220 as hand-written workflows but never registered in `FLOW_INPUTS`, so they ran with an empty flow input and `OpenAiImageGeneration` rejected the empty prompt.
- `test_write_image_metadata` was fixed by hand in #251 to set the required `metadata` parameter, but that edit lived in a generated file and was silently clobbered when #345 reran `scripts/generate_test_workflows.py`.

Both fixes now live in the generator so they survive regeneration: the wide/tall entries go through `MANUAL_FLOW_INPUTS`, and the `single_image` template gains a `set_params` config key that emits `SetParameterValueRequest` calls (used to preset `metadata` on `WriteImageMetadataNode`). The touched workflow files are regenerated output.

Verified locally: all three previously failing tests pass (`write_image_metadata` runs fully local, the two OpenAI ones against the live proxy).

Fixes integration test portion of #409

### DIFF
--- a/scripts/generate_test_workflows.py
+++ b/scripts/generate_test_workflows.py
@@ -159,6 +159,8 @@ ADVANCED_CONFIGS = [
         "image_in": "input_image",
         "output_param": "output_image",
         "suffix": "write_image_metadata",
+        # WriteImageMetadataNode rejects empty metadata, so the workflow must set it.
+        "set_params": {"metadata": {"author": "griptape-nodes-tests", "description": "integration test"}},
     },
     # single_image — API proxy nodes
     {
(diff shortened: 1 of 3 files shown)


scripts/generate_test_workflows.py
feltech, line 426: [consider][minor] Subtle difference here, at first I thought `_SET_PARAM_VALUE ` was a duplicate. Ideally remove `_SET_PARAM` and just use `_SET_PARAM_VALUE`  with `repr(value)` in `_set_param()`.

feltech (approved):
LGTM. Re-ran the previously failing workflows locally and they all pass.

Just one minor consideration, but non-blocking.

scripts/generate_test_workflows.py
feltech: Hmm, OK, didn't realise there was a "merge as soon as reviewed" trigger.  I guess I'd better be careful when approving PRs with `[consider]` comments in future.

PR #419, opened 2026-07-08
feat: add ResolveMacroPath node

shhlife:
## Summary

- Adds `BaseResolveMacroPath` with shared `_extract_path_string` and `_resolve_one` helpers
- Adds `ResolveMacroPath`: single path input (`ParameterString`) → single `resolved_path` output
- Adds `ResolveMacroPaths`: `ParameterList` input → `resolved_paths` list output
- Both are `SuccessFailureNode` — malformed or unresolvable macros surface as a routable failure rather than a silent passthrough
- Accepts any artifact type (plain strings, `UrlArtifact`, `ImageUrlArtifact`, etc.)

Closes #286

## Test plan

- [ ] Wire `Select From Project` → `ResolveMacroPath` and confirm the output is an absolute path
- [ ] Wire an `ImageUrlArtifact` source → `ResolveMacroPath` and confirm it resolves correctly
- [ ] Pass a malformed macro (e.g. `{unclosed`) and confirm the failure edge fires with a useful message
- [ ] Wire multiple paths into `ResolveMacroPaths` and confirm `resolved_paths` contains all of them in order
- [ ] Pass a plain absolute path and confirm it passes through unchanged
- [ ] Confirm both nodes appear in the palette under the **files** category

### DIFF
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -6038,8 +6038,23 @@
       "file_path": "griptape_nodes_library/files/resolve_macro_path.py",
       "metadata": {
         "category": "files",
-        "description": "Resolve one or more macro paths (e.g. {inputs}/file.txt) to absolute filesystem paths.",
+        "description": "Resolve a macro path (e.g. {inputs}/file.txt) to an absolute filesystem path.",
         "display_name": "Resolve Macro Path",
+        "icon": "FolderSearch",
+        "tags": [
+          "file",
+          "macro",
+          "path"
+        ]
+      }
+    },
+    {
+      "class_name": "ResolveMacroPaths",
+      "file_path": "griptape_nodes_library/files/resolve_macro_paths.py",
+      "metadata": {
+        "category": "files",
+        "description": "Resolve a list of macro paths (e.g. {inputs}/file.txt) to absolute filesystem paths.",
+        "display_name": "Resolve Macro Paths",
         "icon": "FolderSearch",
         "tags": [
           "file",
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -6038,7 +6038,7 @@
       "file_path": "griptape_nodes_library/files/resolve_macro_path.py",
       "metadata": {
         "category": "files",
-        "description": "Resolve a macro path (e.g. {inputs}/file.txt) to an absolute filesystem path.",
+        "description": "Resolve a macro path to an absolute filesystem path (e.g. {inputs}/file.txt → /home/user/project/inputs/file.txt).",
         "display_name": "Resolve Macro Path",
         "icon": "FolderSearch",
         "tags": [
@@ -6053,8 +6053,8 @@
       "file_path": "griptape_nodes_library/files/resolve_macro_paths.py",
       "metadata": {
         "category": "files",
-        "description": "Resolve a list of macro paths (e.g. {inputs}/file.txt) to absolute filesystem paths.",
-        "display_name": "Resolve Macro Paths",
+        "description": "Resolve a list of macro paths to absolute filesystem paths (e.g. [{inputs}/file.txt, {outputs}/image.png] → [/home/user/project/inputs/file.txt, /home/user/project/outputs/image.png]).",
+        "display_name": "Resolve Macro Path List",
         "icon": "FolderSearch",
         "tags": [
           "file",
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -6034,6 +6034,21 @@
       }
     },
     {
+      "class_name": "ResolveMacroPath",
+      "file_path": "griptape_nodes_library/files/resolve_macro_path.py",
+      "metadata": {
+        "category": "files",
+        "description": "Resolve one or more macro paths (e.g. {inputs}/file.txt) to absolute filesystem paths.",
+        "display_name": "Resolve Macro Path",
+        "icon": "FolderSearch",
+        "tags": [
+          "file",
+          "macro",
+          "path"
+        ]
+      }
+    },
+    {
       "class_name": "FilePathComponents",
       "file_path": "griptape_nodes_library/files/path.py",
       "metadata": {
--- a/griptape_nodes_library/files/resolve_macro_path.py
+++ b/griptape_nodes_library/files/resolve_macro_path.py
@@ -1,0 +1,92 @@
+from typing import Any
+
+from griptape.artifacts import UrlArtifact
+from griptape_nodes.common.macro_parser import MacroSyntaxError, ParsedMacro
+from griptape_nodes.exe_types.core_types import Parameter, ParameterList, ParameterMode
+from griptape_nodes.exe_types.node_types import DataNode
+from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
+from griptape_nodes.retained_mode.events.project_events import (
+    GetPathForMacroRequest,
+    GetPathForMacroResultSuccess,
+)
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+
+class ResolveMacroPath(DataNode):
+    """Resolve one or more macro paths (e.g. {inputs}/file.txt) to absolute filesystem paths."""
+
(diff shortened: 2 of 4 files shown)


griptape_nodes_library/files/resolve_macro_path.py
```diff
@@ -0,0 +1,92 @@
...
+            Parameter(
+                name="resolved_paths",
+                type="list",
+                allowed_modes={ParameterMode.OUTPUT},
+                default_value=[],
+                tooltip="All resolved absolute paths, in the same order as the inputs.",
+            )
+        )
+
+    def _extract_path_string(self, value: Any) -> str | None:
```
SavagePencil: nit: usually do privates after publics

griptape_nodes_library/files/resolve_macro_path.py
```diff
@@ -0,0 +1,92 @@
...
+from griptape_nodes.exe_types.node_types import DataNode
+from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
+from griptape_nodes.retained_mode.events.project_events import (
+    GetPathForMacroRequest,
+    GetPathForMacroResultSuccess,
+)
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+
+class ResolveMacroPath(DataNode):
```
SavagePencil: Success/Failure?

griptape_nodes_library/files/resolve_macro_path.py
```diff
@@ -0,0 +1,92 @@
...
+    def _resolve_one(self, path_str: str) -> str:
+        """Resolve a single macro path to an absolute path, returning the input as-is on failure."""
+        try:
+            result = GriptapeNodes.handle_request(
+                GetPathForMacroRequest(parsed_macro=ParsedMacro(path_str), variables={})
+            )
+            if isinstance(result, GetPathForMacroResultSuccess):
+                return str(result.resolved_path)
+        except MacroSyntaxError:
+            pass
```
SavagePencil: shouldn't this be a bad thing (tm)?

griptape_nodes_library/files/resolve_macro_path.py
```diff
@@ -0,0 +1,92 @@
...
+        return cleaned if cleaned else None
+
+    def _resolve_one(self, path_str: str) -> str:
+        """Resolve a single macro path to an absolute path, returning the input as-is on failure."""
+        try:
+            result = GriptapeNodes.handle_request(
+                GetPathForMacroRequest(parsed_macro=ParsedMacro(path_str), variables={})
+            )
+            if isinstance(result, GetPathForMacroResultSuccess):
+                return str(result.resolved_path)
```
SavagePencil: if we fail on this, should we at least warn?

SavagePencil (changes requested):

griptape_nodes_library/files/resolve_macro_path.py
```diff
@@ -0,0 +1,92 @@
...
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+
+class ResolveMacroPath(DataNode):
+    """Resolve one or more macro paths (e.g. {inputs}/file.txt) to absolute filesystem paths."""
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+
+        self.paths_input = ParameterList(
```
SavagePencil: any value in making this:
1. Accept EITHER a list OR a single str
2. Split into a single str one vs. a list one?

griptape_nodes_library/files/resolve_macro_path.py
shhlife: @SavagePencil I thought about that - but we're not actually checking files on disk.. just resolving whatever macro is put in, so thought maybe failure wasn't necessary - what do you think?

griptape_nodes_library/files/resolve_macro_path.py
shhlife: @SavagePencil it does work with either a list or a single string.. you can just connect whichever you want. If you don't want to do a list, you don't need to.. if you want you can take a whole list and it just handles it. you can mix & match as well.

I was trying to make it so things wouldn't break downstream if you changed inputs

griptape_nodes_library/files/resolve_macro_path.py
shhlife: sure thing, can do!

griptape_nodes_library/files/resolve_macro_path.py
shhlife: yeah, probs :)

griptape_nodes_library/files/resolve_macro_path.py
SavagePencil: but you are still stuck with a list output, or does that "work" as a string too?

shhlife: that's why there are two outputs.. the first one is always a string - either the only string or the first item in the list. The second is always a list

shhlife: I thought about making it more dynamic, but again - didn't want to break downstream connections

griptape_nodes_library/files/resolve_macro_path.py
SavagePencil: thinking here is what if you give it:
1. A malformed Macro string
2. A well-formed Macro string, but one that can't resolve

griptape_nodes_library/files/resolve_macro_path.py
SavagePencil: nm, you can either:
1. Run it once and get the list entries with outputs
2. For programmatic behavior, up to you to get item from the list output

griptape_nodes_library/files/resolve_macro_path.py
shhlife: all good - will switch!

griptape_nodes_library/files/resolve_macro_path.py
shhlife: gonna make it 2 nodes :)

griptape_nodes_library.json
```diff
@@ -6033,6 +6033,36 @@
...
+        ]
+      }
+    },
+    {
+      "class_name": "ResolveMacroPaths",
+      "file_path": "griptape_nodes_library/files/resolve_macro_paths.py",
+      "metadata": {
+        "category": "files",
+        "description": "Resolve a list of macro paths (e.g. {inputs}/file.txt) to absolute filesystem paths.",
+        "display_name": "Resolve Macro Paths",
```
SavagePencil: could we make the display name a little clearer that it's taking a List? Real easy to miss the `s` at the end.

griptape_nodes_library.json
```diff
@@ -6033,6 +6033,36 @@
...
+          "path"
+        ]
+      }
+    },
+    {
+      "class_name": "ResolveMacroPaths",
+      "file_path": "griptape_nodes_library/files/resolve_macro_paths.py",
+      "metadata": {
+        "category": "files",
+        "description": "Resolve a list of macro paths (e.g. {inputs}/file.txt) to absolute filesystem paths.",
```
SavagePencil: Can we make the description:
1. Depict how it handles a LIST of inputs, and give example real world outputs
2. Close our parentheses like good little students

griptape_nodes_library.json
```diff
@@ -6033,6 +6033,36 @@
         ]
       }
     },
+    {
+      "class_name": "ResolveMacroPath",
+      "file_path": "griptape_nodes_library/files/resolve_macro_path.py",
+      "metadata": {
+        "category": "files",
+        "description": "Resolve a macro path (e.g. {inputs}/file.txt) to an absolute filesystem path.",
```
SavagePencil: Can we make the description:

1. Depict how it converts an actual macro input and give example real world output for it
2. Close our parentheses like good little students

SavagePencil (changes requested):

SavagePencil (approved):

PR #421, opened 2026-07-08
feat: add SetVariableSubstitution node

shhlife:
## Summary

Closes #417.

- Adds a new **Set Variable Substitution** node (`griptape_nodes_library/variables/set_variable_substitution.py`) that enables or disables inline `{VARIABLE_NAME}` token substitution for the current workflow
- The node issues `SetVariableSubstitutionEnabledRequest(enabled=...)` under the hood, replacing the need for users to hand-wire a generic Engine Request node and know the internal request name
- Registered in `griptape_nodes_library.json` alongside the existing variable nodes (`Create Variable`, `Set Variable`, `Get Variable`, `Has Variable`)

The node is a `SuccessFailureNode` with a single boolean `enabled` parameter (default `True`). The setting is workflow-scoped and baked into the saved workflow file on save.

## Test plan

- [ ] Drop a **Set Variable Substitution** node into a workflow
- [ ] Verify it appears in the Variables category in the node palette
- [ ] Set `enabled = True`, run — confirm status shows "Variable substitution enabled."
- [ ] Set `enabled = False`, run — confirm status shows "Variable substitution disabled."
- [ ] Wire a text node with a `{MY_VAR}` token; confirm substitution is applied / suppressed based on the flag
- [ ] Confirm the failure output edge fires when run outside a valid workflow context

### DIFF
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -5978,6 +5978,22 @@
       }
     },
     {
+      "class_name": "SetVariableSubstitution",
+      "file_path": "griptape_nodes_library/variables/set_variable_substitution.py",
+      "metadata": {
+        "category": "variables",
+        "description": "Enable or disable inline variable substitution ({VARIABLE_NAME} token replacement) for this workflow. The setting is scoped to the current workflow and saved with it.",
+        "display_name": "Set Variable Substitution",
+        "icon": "Replace",
+        "group": "edit",
+        "tags": [
+          "variable",
+          "workflow",
+          "substitution"
+        ]
+      }
+    },
+    {
       "class_name": "FileOutputSettings",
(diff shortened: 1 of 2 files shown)


griptape_nodes_library/files/resolve_macro_path.py
SavagePencil, line 1: I think you need to rebase after the other PR goes in, so that this one is JUST the set var substitution node and not all of them.

SavagePencil (changes requested):

griptape_nodes_library/files/resolve_macro_path.py
shhlife: yikes, good catch! fixing

SavagePencil (approved):

PR #441, opened 2026-07-09
feat(describe-image): provider selection + list image input

shhlife:
Closes #439

## Summary

- **Provider selection**: adds a `model_provider` dropdown (mirrors the Agent node pattern) with a refresh button and dynamic model list via `ListAgentProvidersRequest` / `ListProviderModelsRequest`. Non-GTC providers use `GtOpenAiChatPromptDriver` and forward provider credentials in the agent wrapper so downstream nodes can restore the driver.
- **List image input**: flattens nested lists before processing — when a `List[ImageUrlArtifact]` is connected to a `ParameterList` child it arrives as a list-inside-a-list; the node now handles that transparently.
- **String path/URL fix**: string inputs to `images` are now wrapped in `ImageUrlArtifact` and loaded as image bytes rather than passed as text to the model (previously the LLM would see the file path string instead of the image).
- **Validation fix**: replaces the broken `validate_before_workflow_run` (which checked a non-existent `driver` param) with `_uses_griptape_cloud_driver()`, matching the Agent node's approach so non-GTC runs aren't incorrectly blocked.

## Test plan

- [ ] Connect a `SelectFromGrid` (or any node producing `List[ImageUrlArtifact]`) directly to `DescribeImage.images` — node should describe all images
- [ ] Switch provider dropdown to a non-GTC provider (e.g. Ollama) — model list should update and the node should run without requiring `GT_CLOUD_API_KEY`
- [ ] Wire `DescribeImage` into an agent chain and verify the downstream agent sees the description in memory (not base64 bytes)
- [ ] Existing single-image and GTC model flows unchanged

### DIFF
--- a/griptape_nodes_library/image/describe_image.py
+++ b/griptape_nodes_library/image/describe_image.py
@@ -241,7 +241,8 @@ class DescribeImage(ControlNode):
         if provider_name == "griptape_cloud":
             # Use a curated vision-only subset rather than the full GTC model list.
             models = GTC_VISION_MODEL_CHOICES
-            new_data = MODEL_CHOICES_ARGS
+            vision_names = set(GTC_VISION_MODEL_CHOICES)
+            new_data = [entry for entry in MODEL_CHOICES_ARGS if entry["name"] in vision_names]
         else:
             models = self._fetch_models_for_provider(provider_name)
             new_data = [{"name": m, "icon": "", "args": {}} for m in models]
--- a/griptape_nodes_library/image/describe_image.py
+++ b/griptape_nodes_library/image/describe_image.py
@@ -19,6 +19,13 @@ from griptape_nodes.exe_types.param_components.model_access_component import Mod
 from griptape_nodes.exe_types.param_types.parameter_bool import ParameterBool
 from griptape_nodes.exe_types.param_types.parameter_json import ParameterJson
 from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
+from griptape_nodes.retained_mode.events.agent_events import (
+    ListAgentProvidersRequest,
+    ListAgentProvidersResultSuccess,
+    ListProviderModelsRequest,
+    ListProviderModelsResultSuccess,
+    ProviderConfig,
+)
 from griptape_nodes.retained_mode.events.connection_events import CreateConnectionRequest, DeleteConnectionRequest
 from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes, logger
 from griptape_nodes.traits.button import Button, ButtonDetailsMessagePayload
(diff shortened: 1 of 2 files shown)


griptape_nodes_library/image/describe_image.py
```diff
@@ -28,10 +37,38 @@
...
+        ListProviderModelsRequest,  # pyright: ignore[reportAttributeAccessIssue]
+        ListProviderModelsResultSuccess,  # pyright: ignore[reportAttributeAccessIssue]
+        ProviderConfig,  # pyright: ignore[reportAttributeAccessIssue]
+    )
+
+_AGENT_PROVIDERS_AVAILABLE: bool = False
+_GRIPTAPE_CLOUD_PROVIDER: "ProviderConfig" = cast("ProviderConfig", None)
+
+try:
+    from griptape_nodes.retained_mode.events.agent_events import (
```
cjkindel: This has a very strong smell, what is happenign w/ the type_checking and then optimistic imports?

shhlife: Yeah, this is currently the same pattern in `agent.py` - it was done to ensure the agent still works with older engines.

There's another task to refactor this into a shared component https://github.com/griptape-ai/griptape-nodes-library-standard/issues/442

and then a later task to refactor agents to match what we're doing with the sidebar agent - but this gets things working now & shouldn't break older engines.

cjkindel: But we don't need to support older engines, the library requires engine v90 or newer (unless this is a very-recent engine change)

shhlife: happy to pull it out - just didn't want to break anything right before Siggraph

griptape_nodes_library/image/describe_image.py
```diff
@@ -148,6 +208,90 @@ def __init__(self, **kwargs) -> None:
...
+                return cast(ListProviderModelsResultSuccess, result).models or GTC_VISION_MODEL_CHOICES
+        except Exception:
+            pass
+        return GTC_VISION_MODEL_CHOICES
+
+    def _update_model_choices_for_provider(self, provider_name: str) -> None:
+        if provider_name == "griptape_cloud":
+            # Use a curated vision-only subset rather than the full GTC model list.
+            models = GTC_VISION_MODEL_CHOICES
+            new_data = MODEL_CHOICES_ARGS
```
cjkindel: AI Assisted Review:

> **GTC provider branch shows non-vision models in the dropdown.**
> In _update_model_choices_for_provider, the griptape_cloud branch sets the Options choices to the curated 15-model GTC_VISION_MODEL_CHOICES but sets the data ui_option (which the frontend actually renders the dropdown rows from) to the full 28-model MODEL_CHOICES_ARGS. When a user switches the provider dropdown back to griptape_cloud (fires after_value_set → _update_model_choices_for_provider), the model dropdown lists 13 non-vision models that DescribeImage should not offer (deepseek-v3, deepseek.r1-v1, llama3-*, gpt-5-nano, o1, o3-mini, the *-flash* Gemini variants, gpt-5.2-chat, claude-haiku-4-5). The mirrored code in agent.py is correct only because agent.py's choices are the full MODEL_CHOICES; here they diverge. Fix: build new_data from GTC_VISION_MODEL_CHOICES (filter MODEL_CHOICES_ARGS to the vision subset) so data matches choices.

cjkindel (approved):
Only approve b/c SIGGRAPH. The import mess **needs** to be cleaned up afterwards

PR #444, opened 2026-07-09
fix: use native OllamaPromptDriver for Ollama providers (tool calling)

shhlife:
Closes #443

## Summary

- When the Ollama provider is selected, the Agent node now uses `OllamaPromptDriver` (native Ollama client) instead of `OpenAiChatPromptDriver` (OpenAI-compatible endpoint). Ollama's OpenAI-compat endpoint does not correctly handle streaming tool calls (https://github.com/ollama/ollama/issues/9092), causing silent blank output whenever tools are connected.
- The same fix is applied in `restore_provider_driver` in `agent_utils.py`, so all other nodes that accept an agent as input (`mcp_task`, `summarize_agent_memory`, `create_image`) are covered by the same change.
- The provider wrapper dict now includes `"type"` so downstream nodes in a chain can identify the provider and rebuild the correct driver.
- Older agent wires without `"type"` fall back to the OpenAI-compat driver, preserving backwards compatibility with saved workflows.

image

## Test plan

- [ ] Agent node with Ollama provider + DateTime (or other) tool connected → agent responds and tool is called
- [ ] Agent chained through multiple Agent nodes with Ollama + tool → tool calls work end-to-end
- [ ] Agent node with LM Studio / other non-Ollama provider + tool → still works (OpenAI-compat path unchanged)
- [ ] Agent node with Griptape Cloud provider → unaffected

### DIFF
--- a/griptape_nodes_library/agents/agent.py
+++ b/griptape_nodes_library/agents/agent.py
@@ -24,6 +24,7 @@ from griptape.events import (
 from griptape.memory.structure import ConversationMemory, Run
 from griptape.structures import Structure
 from griptape.tasks import PromptTask
+from griptape_nodes.drivers.cloud_models import ProviderID
 from griptape_nodes.exe_types.core_types import (
     NodeMessageResult,
     Parameter,
@@ -1011,7 +1012,7 @@ class Agent(ControlNode):
             # Wrappers from older versions lack "type"; those fall through to the OpenAI-compat driver.
             if incoming_provider:
                 incoming_base_url = incoming_provider.get("base_url", "")
-                if incoming_provider.get("type") == "ollama":
+                if incoming_provider.get("type") == ProviderID.OLLAMA:
                     rebuilt_driver = GtOllamaPromptDriver(
                         model=cast(PromptTask, agent.tasks[0]).prompt_driver.model,
                         host=ollama_host_from_base_url(incoming_base_url),
@@ -1048,7 +1049,7 @@ class Agent(ControlNode):
                 provider_config = next((p for p in providers if p.name == provider_name), _GRIPTAPE_CLOUD_PROVIDER)
                 base_url = provider_config.base_url or ""
                 api_key = self._resolve_provider_api_key(provider_config)
-                if provider_config.type == "ollama":
+                if provider_config.type == ProviderID.OLLAMA:
                     prompt_driver = GtOllamaPromptDriver(
                         model=model_input,
                         host=ollama_host_from_base_url(base_url),
--- a/griptape_nodes_library/agents/agent.py
+++ b/griptape_nodes_library/agents/agent.py
@@ -1021,6 +1021,11 @@ class Agent(ControlNode):
                 incoming_base_url = incoming_provider.get("base_url", "")
                 model = agent.tasks[0].prompt_driver.model
                 if incoming_provider.get("type") == ProviderID.OLLAMA:
+                    # Native Ollama driver is required for tool calling; the OpenAI-compat path produces blank output.
+                    # Trade-off: ollama.Client accepts no api_key, so Ollama instances behind an auth reverse proxy
+                    # cannot have their credentials forwarded here. Use a non-ollama provider type for that setup.
+                    # TODO: remove once griptape exposes headers/api_key on OllamaPromptDriver
+                    #   https://github.com/griptape-ai/griptape/issues/2238
                     rebuilt_driver = GtOllamaPromptDriver(
                         model=model,
                         host=ollama_host_from_base_url(incoming_base_url),
@@ -1058,6 +1063,11 @@ class Agent(ControlNode):
                 base_url = provider_config.base_url or ""
                 api_key = self._resolve_provider_api_key(provider_config)
                 if provider_config.type == ProviderID.OLLAMA:
+                    # Native Ollama driver is required for tool calling; the OpenAI-compat path produces blank output.
+                    # Trade-off: ollama.Client accepts no api_key, so Ollama instances behind an auth reverse proxy
+                    # cannot have their credentials forwarded here. Use a non-ollama provider type for that setup.
+                    # TODO: remove once griptape exposes headers/api_key on OllamaPromptDriver
+                    #   https://github.com/griptape-ai/griptape/issues/2238
                     prompt_driver = GtOllamaPromptDriver(
                         model=model_input,
                         host=ollama_host_from_base_url(base_url),
--- a/griptape_nodes_library/agents/agent.py
+++ b/griptape_nodes_library/agents/agent.py
@@ -12,8 +12,6 @@ from typing import TYPE_CHECKING, Any, cast  # cast used for handle_request narr
 from griptape.artifacts import BaseArtifact, ModelArtifact, TextArtifact
 from griptape.drivers.prompt.base_prompt_driver import BasePromptDriver
 from griptape.drivers.prompt.griptape_cloud import GriptapeCloudPromptDriver
-from griptape.drivers.prompt.ollama import OllamaPromptDriver as GtOllamaPromptDriver
-from griptape.drivers.prompt.openai import OpenAiChatPromptDriver as GtOpenAiChatPromptDriver
 from griptape.events import (
     ActionChunkEvent,
     FinishActionsSubtaskEvent,
@@ -24,7 +22,6 @@ from griptape.events import (
 from griptape.memory.structure import ConversationMemory, Run
 from griptape.structures import Structure
 from griptape.tasks import PromptTask
-from griptape_nodes.drivers.cloud_models import ProviderID
 from griptape_nodes.exe_types.core_types import (
     NodeMessageResult,
     Parameter,
@@ -51,9 +48,9 @@ from griptape_nodes_library.config.prompt.cloud_models import (
     MODEL_CHOICES_ARGS,
 )
 from griptape_nodes_library.utils.agent_utils import (
+    build_prompt_driver,
     build_rulesets_from_configs,
     build_tools,
-    ollama_host_from_base_url,
     ruleset_to_config,
     unwrap_agent,
     wrap_agent,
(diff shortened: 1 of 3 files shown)


griptape_nodes_library/agents/agent.py
```diff
@@ -1006,13 +1009,22 @@ def process(self) -> AsyncResult[Structure]:  # noqa: C901, PLR0915, PLR0912
...
-                rebuilt_driver = GtOpenAiChatPromptDriver(
-                    model=cast(PromptTask, agent.tasks[0]).prompt_driver.model,
-                    base_url=incoming_provider.get("base_url", ""),
-                    api_key=incoming_provider.get("api_key") or "not-needed",
-                    stream=True,
-                )
+                incoming_base_url = incoming_provider.get("base_url", "")
+                if incoming_provider.get("type") == ProviderID.OLLAMA:
+                    rebuilt_driver = GtOllamaPromptDriver(
+                        model=cast(PromptTask, agent.tasks[0]).prompt_driver.model,
```
cjkindel: If `agent.tasks[0]` is not a PromptTask it will get the default (GTC sonnet 4.6). This seems like a dangerous/undesired leak

griptape_nodes_library/utils/agent_utils.py
```diff
@@ -60,34 +66,55 @@ def unwrap_agent(value: dict) -> tuple[dict, list, list]:
...
-    rebuilt = OpenAiChatPromptDriver(
-        model=model,
-        base_url=provider.get("base_url", ""),
-        api_key=provider.get("api_key") or "not-needed",
-        stream=True,
-    )
+    base_url = provider.get("base_url", "")
+
+    if provider.get("type") == ProviderID.OLLAMA:
+        rebuilt = OllamaPromptDriver(model=model, host=ollama_host_from_base_url(base_url), stream=True)
```
cjkindel: AI Assisted Comment:

> **api_key is silently dropped on the Ollama path. (PLAUSIBLE)**
> The` OpenAI-compat path forwards api_key; OllamaPromptDriver/ollama.Client(host=...) accepts no key. Catalog Ollama is NO_KEY_REQUIRED so the standard config carries no real credential, but a user who hand-configured an ollama-type provider with api_key_secret_name (Ollama behind an auth reverse proxy) that previously worked will now get 401/403. Same drop at [agent.py:1016](vscode-webview://0c5bimdqi86ou9obd49pmqt5lqucfalvknrsanmog52ml4hisai8/griptape_nodes_library/agents/agent.py#L1016) and [agent.py:1053](vscode-webview://0c5bimdqi86ou9obd49pmqt5lqucfalvknrsanmog52ml4hisai8/griptape_nodes_library/agents/agent.py#L1053).

cjkindel (changes requested):

griptape_nodes_library/agents/agent.py
```diff
@@ -1006,13 +1009,22 @@ def process(self) -> AsyncResult[Structure]:  # noqa: C901, PLR0915, PLR0912
...
+            # Wrappers from older versions lack "type"; those fall through to the OpenAI-compat driver.
             if incoming_provider:
-                rebuilt_driver = GtOpenAiChatPromptDriver(
-                    model=cast(PromptTask, agent.tasks[0]).prompt_driver.model,
-                    base_url=incoming_provider.get("base_url", ""),
-                    api_key=incoming_provider.get("api_key") or "not-needed",
-                    stream=True,
-                )
+                incoming_base_url = incoming_provider.get("base_url", "")
+                if incoming_provider.get("type") == ProviderID.OLLAMA:
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding.

**Finding: provider→driver selection is duplicated across three call sites.**

The same `if type == OLLAMA → OllamaPromptDriver else → OpenAiChatPromptDriver` branch now lives in three places, each independently threading `model` / `base_url` / `api_key` and repeating the host conversion: here in the incoming-provider rebuild, again in the fresh-provider path, and a third time in `restore_provider_driver`.

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/4fb4b10b48d5fee11a24e2241305ff78e0e34217/griptape_nodes_library/agents/agent.py#L1052-L1064

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/4fb4b10b48d5fee11a24e2241305ff78e0e34217/griptape_nodes_library/utils/agent_utils.py#L109-L117

**Impact:** This is the source of the "why is there Ollama-specific code in an otherwise generic agent?" smell. Adding another special-cased provider, or changing driver kwargs, means editing three spots and risking drift between them.

**Suggested fix:** Extract one factory in `agent_utils.py`, e.g. `build_prompt_driver(*, provider_type, model, base_url, api_key=None, stream=True) -> BasePromptDriver`, and call it from all three sites. Keep it a 2-way `if/else` (Ollama vs. OpenAI-compat default) rather than `match`/`case` with a wildcard `raise` — the fall-through to the OpenAI-compat driver for unknown/missing types is intentional here, so raising on the default would be wrong. Once extracted, the agent stops knowing anything about Ollama and the special case lives in exactly one place.

griptape_nodes_library/utils/agent_utils.py
```diff
@@ -60,34 +66,55 @@ def unwrap_agent(value: dict) -> tuple[dict, list, list]:
     return value, [], []
 
 
+def ollama_host_from_base_url(base_url: str) -> str | None:
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding.

**Finding: the root cause is undocumented and this helper is a public util when it's really an implementation detail.**

The conversion itself is genuinely necessary: provider configs store the OpenAI-compat URL (`http://localhost:11434/v1`), and the `ollama` client's `_parse_host` *preserves* path suffixes, so passing `.../v1` as `host` would build a broken `http://localhost:11434/v1/api/chat`. So the `/v1` strip has to happen somewhere. Two issues with how it lands:

1. **The real reason for the Ollama branch isn't captured anywhere in the code.** The surrounding comments say "pick the right driver" / "Ollama native or OpenAI-compatible," which is exactly what leaves a reader asking *why* Ollama is special. The concrete reason: Ollama's OpenAI-compat `/v1` endpoint cannot stream tool calls — with `tools` + `stream=True` it collapses the stream / drops `tool_calls` ([ollama/ollama#9084](https://github.com/ollama/ollama/issues/9084), [#8887](https://github.com/ollama/ollama/issues/8887)). This library hardcodes `stream=True`, so it lands on the broken path. griptape's native `OllamaPromptDriver` (`use_native_tools=True`) hits `/api/chat` and handles tool-call streaming. That is the one-line justification for the branch existing at all.

2. **This is exposed as a public helper with exactly one legitimate caller** (building the native driver), and it duplicates a different convention: the dedicated `OllamaPrompt` config node builds its host as `f"{base_url}:{port}"` with no `/v1` at all.

**Impact:** The next reader hits the same "why is Ollama unique?" wall we did, and the loose public util invites reuse of a naive string transform that only handles the `/v1` suffix.

**Suggested fix:** Fold the conversion into the `build_prompt_driver` factory as a private `_ollama_host_from_base_url`, and put the root-cause comment (streaming tool calls, ollama/ollama#9084) directly on the Ollama branch. Mirror the same one-liner in the PR description.

griptape_nodes_library/utils/agent_utils.py
```diff
@@ -60,34 +66,55 @@ def unwrap_agent(value: dict) -> tuple[dict, list, list]:
...
+    """Convert an OpenAI-compat Ollama base_url to the host expected by OllamaPromptDriver.
+
+    Provider configs store the OpenAI-compat URL (e.g. http://localhost:11434/v1).
+    OllamaPromptDriver uses the native Ollama client, which takes just the host
+    without any path suffix (e.g. http://localhost:11434). Passing None is valid
+    and causes ollama.Client to default to http://localhost:11434.
+    """
+    host = base_url.rstrip("/")
+    if host.endswith("/v1"):
```
collindutter, line 81: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding.

**Finding: no unit test for a pure, branch-heavy helper.**

`ollama_host_from_base_url` is a pure function with exactly the kind of edge cases that regress silently: trailing-slash normalization, the `/v1` suffix strip, and the empty-string → `None` case. `tests/unit/utils/` already establishes a table-driven pattern for testing pure utilities (e.g. `test_artifact_path_validator.py`).

**Impact:** Per the project convention ("write unit tests if there is an established pattern"), this helper is currently the untested exception. A future change to the strip logic could break host resolution with no test to catch it.

**Suggested fix:** Add a small table-driven test module covering: `http://localhost:11434/v1` → `http://localhost:11434`, trailing slash `.../v1/` → `.../11434`, a non-`/v1` URL left intact, and `""` → `None`. If the helper moves into the factory as a private function per the other comment, test it through the factory instead.

griptape_nodes_library/agents/agent.py
```diff
@@ -1006,13 +1009,22 @@ def process(self) -> AsyncResult[Structure]:  # noqa: C901, PLR0915, PLR0912
             else:
                 agent.tasks[0].output_schema = pydantic_schema
             # Rebuild the prompt driver for non-GTC providers — griptape strips api_key during serialization.
+            # Wrappers from older versions lack "type"; those fall through to the OpenAI-compat driver.
```
collindutter, line 1016: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and independently verified this finding.

**Finding: workflows saved before the `type` key existed silently reintroduce the tool-calling bug.**

A wrapper produced before `"type"` was added to the provider dict has no `type`, so an Ollama agent falls through to the OpenAI-compat driver on reload — which is exactly the broken `/v1` + streaming + tools path this PR is fixing. The docstring documents the gap honestly:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/4fb4b10b48d5fee11a24e2241305ff78e0e34217/griptape_nodes_library/utils/agent_utils.py#L90-L93

but the failure is silent: the agent loads fine and simply stops calling tools.

**Impact:** Existing saved Ollama workflows keep silently failing tool calls until someone re-runs them, with no signal pointing at the cause.

**Suggested fix:** At minimum, log a warning when we fall through to the OpenAI-compat driver because `type` is missing, so the cause is discoverable. Optionally, infer Ollama when `type` is absent but `base_url` matches the known Ollama default/port — this is another heuristic, so logging-only is the safer default unless auto-healing old workflows is worth it.

collindutter (changes requested):

griptape_nodes_library/agents/agent.py
shhlife: we only hit this once, if a user connects an image_generation node to another agent - doesn't happen often, but we'll warn now and make sure to NOT fall back to GTC.

griptape_nodes_library/utils/agent_utils.py
shhlife: This is something that needs to be fixed in the Griptape library - logged and commented as a followup https://github.com/griptape-ai/griptape/issues/2238

collindutter (approved):

cjkindel (approved):

PR #445, opened 2026-07-09
feat: declare model invocations on the agent and driver paths

collindutter:
Part of #431
image

image
image
image

### DIFF
--- a/griptape_nodes_library/agents/agent.py
+++ b/griptape_nodes_library/agents/agent.py
@@ -63,6 +63,7 @@ from griptape_nodes_library.utils.agent_utils import (
     wrap_agent,
 )
 from griptape_nodes_library.utils.error_utils import try_throw_error
+from griptape_nodes_library.utils.model_invocation import declare_model_invocation_sync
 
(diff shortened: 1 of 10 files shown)


griptape_nodes_library/agents/agent.py
```diff
@@ -1055,6 +1056,27 @@ def process(self) -> AsyncResult[Structure]:  # noqa: C901, PLR0915, PLR0912
...
 
         if prompt and not prompt.isspace():
+            # Declare the model that will actually run, taken from the node's own model
+            # parameter where possible: a dropdown selection is the resolved string, a
+            # directly connected driver exposes its `model`. Only a connected Agent keeps
+            # the model inside its restored task, so fall back to that there. The util
+            # resolves the provider model id to its stable catalog key (via the node's
+            # model_usage) before declaring. Declare before the network call below so a
+            # denied invocation fails closed rather than reaching the provider.
+            if isinstance(model_input, str):
```
cjkindel: I think this will still evaluate true if the user had previously set the model dropdown value. It gets hidden, but I don't think the value actually gets cleared when agent dict is connected

cjkindel (changes requested):

cjkindel (approved):
cursed light mode screenshots

PR #446, opened 2026-07-10
Macro Node

SmoolPerson:
This node allows the user to directly interact with the macro system in the engine without needing to call the functions from source code. Users can add their template string, and pass in a dictionary of what to replace each template parameter with. Additionally, the code contains error handling which alerts the user if any template string or parameter is malformed.

### DIFF
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -29,8 +29,8 @@
   "metadata": {
     "author": "Griptape, Inc",
     "description": "Default nodes for Griptape Nodes",
-    "library_version": "0.80.1",
-    "engine_version": "0.89.0",
+    "library_version": "0.82.0",
+    "engine_version": "0.90.0",
     "tags": [
       "Griptape",
       "AI"
@@ -192,6 +192,12 @@
                 "display_name": "Veo 3.0 Fast",
                 "family": "Veo",
                 "provider_model_id": "veo-3.0-fast-generate-001",
+                "key_support": "REQUIRES_GRIPTAPE_KEY"
+              },
+              "gtc_gemini_omni_flash": {
+                "display_name": "Gemini Omni Flash",
+                "family": "Gemini Omni",
+                "provider_model_id": "gemini-omni-flash-preview",
                 "key_support": "REQUIRES_GRIPTAPE_KEY"
               }
             }
@@ -706,6 +712,12 @@
                 "family": "Tripo",
                 "provider_model_id": "tripo-text-to-3d",
                 "key_support": "SUPPORTS_CUSTOMER_KEY_OR_GRIPTAPE_KEY"
+              },
+              "gtc_tripo_multiview_to_3d": {
+                "display_name": "Tripo Multiview to 3D",
+                "family": "Tripo",
+                "provider_model_id": "tripo-multiview-to-3d",
+                "key_support": "SUPPORTS_CUSTOMER_KEY_OR_GRIPTAPE_KEY"
               }
             }
           },
@@ -901,12 +913,12 @@
     {
       "name": "CropVideoEditor",
       "path": "widgets/CropVideoEditor/CropVideoEditor.js",
-      "description": "Interactive video crop editor — drag handles on a live video frame with frame scrubber, aspect ratio, resolution, and position presets."
+      "description": "Interactive video crop editor \u2014 drag handles on a live video frame with frame scrubber, aspect ratio, resolution, and position presets."
     },
     {
       "name": "SelectFromGrid",
       "path": "widgets/SelectFromGrid/SelectFromGrid.js",
-      "description": "Interactive grid selector — displays list items as thumbnails (images, video, audio, text) and lets the user click to select one or more."
+      "description": "Interactive grid selector \u2014 displays list items as thumbnails (images, video, audio, text) and lets the user click to select one or more."
     }
   ],
   "categories": [
@@ -2681,6 +2693,34 @@
       }
     },
     {
+      "class_name": "GeminiOmniFlashGeneration",
+      "file_path": "griptape_nodes_library/video/gemini_omni_flash_generation.py",
+      "metadata": {
+        "category": "video",
+        "description": "Generate video using Google Gemini Omni Flash via Griptape Cloud model proxy",
+        "display_name": "Gemini Omni Flash Video Generation",
+        "icon": "Sparkles",
+        "group": "create",
+        "tags": [
+          "video",
+          "generation",
+          "ai",
+          "api",
+          "gemini",
+          "omni",
+          "google"
+        ],
+        "declarations": [
+          {
+            "type": "model_usage",
+            "model_ids": [
+              "gtc_gemini_omni_flash"
+            ]
+          }
+        ]
+      }
+    },
+    {
       "class_name": "WanImageToVideoGeneration",
       "file_path": "griptape_nodes_library/video/wan_image_to_video_generation.py",
       "metadata": {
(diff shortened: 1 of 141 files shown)


griptape_nodes_library/text/resolve_macro_template.py
```diff
@@ -0,0 +1,118 @@
...
+        self.add_parameter(
+            ParameterString(
+                name="template",
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+                default_value="",
+                multiline=True,
+                placeholder_text="Hello, {name}! You have {count} messages.",
+                tooltip=(
+                    "The macro template string. Reference variables with braces "
+                    "(e.g. '{name}'). Optional variables use '{name?}'."
```
SavagePencil: `"The macro template string. The macro language is described here: https://docs.griptapenodes.com/en/stable/guides/projects/macros/"`

griptape_nodes_library/text/resolve_macro_template.py
```diff
@@ -0,0 +1,118 @@
...
+                ),
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="variables",
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+                type="dict",
```
SavagePencil: for some reason this isn't showing up as a nice dict table. I'm researching; it should for a PROPERTY-mode parameter.

griptape_nodes_library/text/resolve_macro_template.py
```diff
@@ -0,0 +1,118 @@
...
+        variables = self._coerce_variables(raw_variables)
+
+        try:
+            parsed = ParsedMacro(template=template)
+        except MacroSyntaxError as err:
+            self._report_failure(f"Failed to parse template. {err}", exception=err)
+            return
+
+        try:
+            resolved = parsed.resolve(variables, GriptapeNodes.SecretsManager())
```
SavagePencil: suggest we remove passing the SecretsManager for two reasons:
a) reduce the ease of exfiltration
b) where this node is executing may not be where the SecretsManager is in a "worker" configuration (library running in a separate process)

SavagePencil (changes requested):

griptape_nodes_library/text/resolve_macro_template.py
SmoolPerson: Great idea, I could have honestly passed that right into claude instead of it scouring the codebase for macro docs.

griptape_nodes_library/text/resolve_macro_template.py
SavagePencil: Engine task to remove the SecretsManager here: https://github.com/griptape-ai/griptape-nodes-engine/issues/5186

griptape_nodes_library/text/resolve_macro_template.py
SavagePencil: Found the root cause — it's the `default_value={}`. The editor's dict widget builds one row per entry of the current value, and only seeds a blank key/value row when the value is *unset* (`None`). An empty dict is "present but has zero entries," so you get the label with no rows — and there's no add-row button, so it stays that way.

Fix on your side: change to `default_value=None` (renders one blank row), or seed a starter pair the way the existing **Key Value Pair** node does (`default_value={"key": "value"}`).

While you're in here: consider using `ParameterDict` from `griptape_nodes.exe_types.param_types.parameter_dict` instead of a raw `Parameter(type="dict")` — it's the engine's dict-parameter helper (matches the `ParameterString` you're already using) and adds sensible input coercion for free.

(The widget arguably should treat `{}` the same as unset — we'll take that on the editor side separately. No need to wait on it.)

griptape_nodes_library/text/resolve_macro_template.py
SavagePencil: Adding specifics: the engine-side fix is now up — griptape-ai/griptape-nodes-engine#5187 (issue griptape-ai/griptape-nodes-engine#5186) makes `secrets_manager` optional with `None` as the default.

Once that lands and this library's `engine_version` picks it up, the fix here is simply:

```python
resolved = parsed.resolve(variables)
```

— no SecretsManager import, no `GriptapeNodes.SecretsManager()` call.

Behavior note worth reflecting in the tooltip/docs: with no secrets access, a variable value starting with `$` is treated as a secret reference you're not allowed to see. If the variable is **required**, resolution fails with a clear "no secrets access is available" error; if it's **optional**, it's skipped, same as an unprovided optional. So values like `"$100"` will error rather than render — that's the deliberate trade-off for keeping secrets unreachable from this node's output.

griptape_nodes_library/text/resolve_macro_template.py
```diff
@@ -0,0 +1,239 @@
...
+            )
+            return
+
+        self._clear_execution_status()
+
+        template: str = self.get_parameter_value("template")
+        raw_variables = self.get_parameter_value("variables")
+        source = VariableSource(self.get_parameter_value("variable_source"))
+
+        coerced = self._coerce_variables(raw_variables)
```
SavagePencil, line 150: A one-line helper method that's only called in one spot usually doesn't add a ton of value, IMO

griptape_nodes_library.json
SavagePencil, line 1: I believe you need to bump the required engine version for this library to `0.95.0`

griptape_nodes_library/variables/variable_utils.py
```diff
@@ -171,6 +171,48 @@ def has_variable(node_name: str, variable_name: str, scope: "VariableScope") ->
     return result.exists
 
 
+def get_variables(node_name: str, names: list[str], scope: "VariableScope") -> dict[str, Any]:
```
SavagePencil, line 174: is there no way to import the actual StrEnum, or do we have to do the silly string thing for `VariableScope`?

griptape_nodes_library/text/resolve_macro_template.py
```diff
@@ -0,0 +1,239 @@
...
+        return {str(key): self._normalize_variable_value(value) for key, value in raw_variables.items()}
+
+    def _normalize_variable_value(self, value: Any) -> str | int:
+        """Coerce a single value to what ParsedMacro.resolve accepts (str or int)."""
+        # bool is a subclass of int; render it as "True"/"False" instead of 1/0.
+        if isinstance(value, bool):
+            return str(value)
+        if isinstance(value, (str, int)):
+            return value
+        return str(value)
```
SavagePencil, line 232: this has a bad smell to it. Are there:
a) other places where we do this or something like it already
b) cases we need to guard around with a try/catch

griptape_nodes_library/text/resolve_macro_template.py
```diff
@@ -0,0 +1,239 @@
...
+
+        variable_values = self._get_normalized_variables(list(template_names), scope)
+        return {**coerced, **variable_values}
+
+    def _get_normalized_variables(self, names: list[str], scope: VariableScope) -> dict[str, str | int]:
+        """Read selected workflow Variables and normalize their values for macro resolution."""
+        if not names:
+            return {}
+
+        variables = get_variables(self.name, names, scope)
```
SavagePencil, line 213: why would we get ALL variables within the scope? Can't we get just the ones that are specified in the macro string?

SavagePencil: Found it; you might try `GetVariablesRequest`; this will tell you which variables resolved and which did not. You can get the list of variables from the ParsedMacro.

SavagePencil (changes requested):

griptape_nodes_library/text/resolve_macro_template.py
SavagePencil: nm, I need to learn how to read

griptape_nodes_library.json
SmoolPerson: Someone already did this to 0.96.0, so I don't think there's a need.

griptape_nodes_library/variables/variable_utils.py
SmoolPerson: Don't know why Claude decided to use quotes there.

SavagePencil: That can happen with circular dependencies or forward declares or other things that I never remember until I run into them.

griptape_nodes_library.json
SavagePencil: Yep, that was probably Jason getting the piece we identified in and released :)

PR #461, opened 2026-07-14
feat: catalog Amazon Bedrock and Ollama providers

collindutter:
Settles the two catalog conventions blocking full `model_usage` coverage (#429, #430), part of griptape-ai/internal#172.

Bedrock is catalogued as its own `amazon_bedrock` provider whose `provider_model_id`s are the literal Bedrock wire ids (region-prefixed inference-profile ids and all), rather than mapping onto the existing `anthropic`/`deepseek`/`meta` vendor providers. That keeps the convention from #338 that `provider_model_id` is the id the API actually receives, at the cost of the same underlying model appearing under two providers; #437 already accepted that trade for Groq and NIM, and the provider `notes` record it. `AmazonBedrockPrompt` declares all 11 models it offers.

Ollama is the local-runtime case the `KeySupport` docs anticipate: models are discovered from the user's local server at runtime, so the provider declares no models and carries provider-level `key_support: NO_KEY_REQUIRED`. `OllamaPrompt` references the whole provider via `model_provider_usage`. A test pins that `resolve_node_models` yields nothing for it, so the catalog stays descriptive rather than feeding the node a model list, matching the decision recorded in #338.

Also adds a manifest-wide validation test that runs the engine's own `LibrarySchema` + `validate_library_declarations` path, so cross-reference breakage (dangling `model_usage` ids, duplicate model keys) fails in CI instead of at library load.

Closes #429. Closes #430.

> [!NOTE]
> Beep boop. This PR description was written by AI.

### DIFF
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -1018,7 +1018,7 @@
           },
           "nvidia_nim": {
             "display_name": "NVIDIA NIM",
-            "notes": "NVIDIA NIM is a serving platform, not a model vendor; the models below (Llama, Gemma, DeepSeek, Mistral, etc.) are served through NVIDIA's infrastructure under NVIDIA's own model ids. Using the serving platform as the provider key here is provisional pending the platform-as-provider convention discussion in issue #429 (Amazon Bedrock).",
+            "notes": "NVIDIA NIM is a serving platform, not a model vendor; the models below (Llama, Gemma, DeepSeek, Mistral, etc.) are served through NVIDIA's infrastructure under NVIDIA's own model ids. Cataloguing the serving platform as the provider key follows the same convention settled for Amazon Bedrock (see issue #429): provider_model_id stays the literal wire id the platform's API expects, even when it duplicates a model also catalogued under its original vendor.",
             "models": {
               "nim_deepseek_v3_1": {
                 "display_name": "DeepSeek V3.1",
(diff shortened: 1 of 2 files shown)


griptape_nodes_library.json
```diff
@@ -1111,6 +1111,84 @@
                 "key_support": "REQUIRES_CUSTOMER_KEY"
               }
             }
+          },
+          "amazon_bedrock": {
+            "display_name": "Amazon Bedrock",
+            "notes": "Amazon Bedrock re-serves models from other vendors (Anthropic, DeepSeek, Meta) plus Amazon's own Titan models, all under Bedrock's own wire ids -- often region-prefixed cross-region inference profile ids (a 'us.' prefix) and sometimes date-suffixed -- which don't match the bare provider_model_ids the same underlying models use under the anthropic, deepseek, and meta providers elsewhere in this catalog. Bedrock is catalogued as its own provider so provider_model_id stays the literal wire id Bedrock's API expects; the same underlying model can therefore appear under two providers with two different provider_model_ids.",
+            "models": {
+              "bedrock_claude_opus_4_7": {
```
cjkindel, line 1119: Little confused on how this list is defined/where it is from. ex: Bedrock definitely supports Opus 4.8

Assuming its coming from amazon_bedrock_prompt, do we have a way to consolidate the two lists (I thought we did from a previous PR)

griptape_nodes_library.json
```diff
@@ -1018,7 +1018,7 @@
           },
           "nvidia_nim": {
             "display_name": "NVIDIA NIM",
-            "notes": "NVIDIA NIM is a serving platform, not a model vendor; the models below (Llama, Gemma, DeepSeek, Mistral, etc.) are served through NVIDIA's infrastructure under NVIDIA's own model ids. Using the serving platform as the provider key here is provisional pending the platform-as-provider convention discussion in issue #429 (Amazon Bedrock).",
+            "notes": "NVIDIA NIM is a serving platform, not a model vendor; the models below (Llama, Gemma, DeepSeek, Mistral, etc.) are served through NVIDIA's infrastructure under NVIDIA's own model ids. Cataloguing the serving platform as the provider key follows the same convention settled for Amazon Bedrock (see issue #429): provider_model_id stays the literal wire id the platform's API expects, even when it duplicates a model also catalogued under its original vendor.",
```
cjkindel, line 1021: nit: Closing issue #429 but the code still references it?

collindutter: Intentional: the reference is a pointer to where the convention was settled (the decision record), not a live dependency on the issue staying open. Happy to drop the parenthetical if it reads as unfinished business, the note stands on its own without it.

griptape_nodes_library.json
collindutter: The list is exactly `AmazonBedrockPrompt`'s `MODEL_CHOICES` (its dropdown), in the same order. The `model_catalog` declares what this library's nodes actually offer, not everything Bedrock supports, so Opus 4.8 is absent because the node doesn't offer it today. Adding it to the node would be a separate change, and the consistency test added here would then force the catalog entry to come along with it.

On consolidation: the two lists are deliberately kept as two lists guarded by `test_model_selection_node_usage_matches_static_choices` (extended here to cover `AmazonBedrockPrompt`). Sourcing the dropdown from the catalog was tried in #338 and backed out, since it makes the catalog load-bearing at runtime, where a manifest error breaks the node instead of a test. Whether to revisit that (including a build-time generation hybrid) is tracked in #448.

cjkindel: Heard on 4.8

And sadly acknowledge 338/448

cjkindel (approved):

griptape_nodes_library.json
collindutter: Didn't mean for AI to respond directly lol. But yeah, that ^

cjkindel: Damn, I thought that one was an authentic (tired of PRs) Collin response. spooky

PR #467, opened 2026-07-15
fix(subflow): disambiguate node<->flow link using UUID

feltech:
Closes #466.

Nodes are grouped into flows. Flows are uniquely identified by their name (distinct from node names). These flow names can change due to de-duplication when loading a workflow file.

In particular the SubflowWorkflowNode node loads a flow and its nodes from an external workflow file, and if the external workflow file happens to have a flow whose name conflicts with another (the parent or another previously loaded flow), then it is renamed. An external workflow file can itself have SubflowWorkflowNode nodes, adding a cascade of renames during the deserialisation process.

Therefore, flow names are unstable identifiers. The SubflowWorkflowNode made the mistake of assuming the flow it is associated with could be stably identified by its name, and so baked in a fixed flow name into the node's metadata.

This could manifest as the subflow not being found, or the wrong subflow being used, infinite recursion, and deletion of the wrong subflow when the node is deleted.

So add a UUID to all new SubflowWorkflowNode nodes, and stamp that UUID on the flow loaded from its associated workflow file. Store the UUID in node/flow metadata so that is survives serialisation. Use the UUID to disambiguate subflow name collisions and search through available flows as a fallback.

Make a best effort to support workflows saved before these UUIDs were introduced, by claiming an available matching subflow by name if the flow is not already claimed by UUID and the flow has an association with the expected workflow file. I.e. equivalent to what we used to do, with the added protection of ensuring two SubflowWorkflowNodes cannot claim the same flow and that the flow came from the correct workflow file.

Update the test_subflow_workflow_node.py e2e workflow test introduced in #460 to have a middle workflow layer that holds 2x SubflowWorkflowNodes that both reference the same "echo" workflow. This is a minimal simulation of the original issue, whilst maintaining coverage of the fix in #460.

Add a duplicate e2e workflow test to the above, but created before this fix was introduced (i.e. so no serialised UUIDs), to test the legacy fallback path. Tag these tests with a `_legacy` suffix.

### DIFF
--- a/griptape_nodes_library/engine/subflow_workflow_node.py
+++ b/griptape_nodes_library/engine/subflow_workflow_node.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from typing import Any
+from uuid import uuid4
 
 from griptape_nodes.exe_types.core_types import Parameter, ParameterMode, ParameterTypeBuiltin
 from griptape_nodes.exe_types.flow import ControlFlow
@@ -10,7 +11,7 @@ from griptape_nodes.retained_mode.events.execution_events import (
     StartLocalSubflowRequest,
     StartLocalSubflowResultFailure,
 )
-from griptape_nodes.retained_mode.events.flow_events import DeleteFlowRequest
+from griptape_nodes.retained_mode.events.flow_events import DeleteFlowRequest, SetFlowMetadataRequest
 from griptape_nodes.retained_mode.events.node_events import GetFlowForNodeRequest, GetFlowForNodeResultSuccess
 from griptape_nodes.retained_mode.events.parameter_events import (
     RemoveParameterFromNodeRequest,
@@ -28,10 +29,30 @@ from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes, logger
 from griptape_nodes.traits.button import Button, ButtonDetailsMessagePayload
 from griptape_nodes.traits.options import Options
 
+# Node-metadata key holding the name of the imported child flow this node last
+# bound to. It is a fast hint only; ``SUBFLOW_OWNER_KEY`` is the authoritative
+# link (see below).
+SUBFLOW_NAME_KEY = "subflow_name"
+
+# Node-metadata key recording which registered workflow the current subflow was
+# imported from, so a reload can tell "already loaded" from "workflow changed".
+SUBFLOW_WORKFLOW_KEY = "_subflow_workflow"
+
+# Metadata key stored on BOTH the Workflow node and its imported child flow. The
+# shared UUID binds a specific node to the specific subflow it imported, so the
+# two stay linked across saves, flow-name de-duplication, and nesting where the
+# recorded ``subflow_name`` may no longer match.
+SUBFLOW_OWNER_KEY = "subflow_owner_uuid"
+
 
 class SubflowWorkflowNode(SuccessFailureNode):
     def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
         super().__init__(name, metadata)
+        # Establish this node's stable owner UUID once, up front, so it exists
+        # before any subflow import needs it. Preserved from saved metadata on
+        # load; generated for brand-new nodes. It binds the node to the specific
+        # child flow it imports.
+        self.metadata.setdefault(SUBFLOW_OWNER_KEY, str(uuid4()))
         result = GriptapeNodes.handle_request(ListCallableWorkflowsRequest())
         if isinstance(result, ListCallableWorkflowsResultSuccess):
             # Add pyright ignore here becuase we know ListCallableWOrkflowsResultSuccess reutrns workflow_names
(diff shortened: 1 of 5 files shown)


griptape_nodes_library/engine/subflow_workflow_node.py
```diff
@@ -28,10 +29,30 @@
...
+
 
 class SubflowWorkflowNode(SuccessFailureNode):
     def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
         super().__init__(name, metadata)
+        # Establish this node's stable owner UUID once, up front, so it exists
+        # before any subflow import needs it. Preserved from saved metadata on
+        # load; generated for brand-new nodes. It binds the node to the specific
+        # child flow it imports.
+        self.metadata.setdefault(SUBFLOW_OWNER_KEY, str(uuid4()))
```
collindutter, line 60: > [!NOTE]

**Finding: the owner UUID is not unique when the same workflow is instantiated more than once.**

`setdefault` only generates a fresh UUID when the key is absent, so it preserves any `subflow_owner_uuid` baked into saved metadata. Because that UUID is written as a literal into the serialized workflow and import-as-referenced re-runs the workflow file, instantiating the same workflow more than once produces multiple nodes that all carry the *same* owner UUID. Each then stamps its imported child flow with that same UUID:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/e83b94952901340de1e61e4aaaaf030e02672b30/griptape_nodes_library/engine/subflow_workflow_node.py#L176-L182

Resolution keys off that UUID. The quick path trusts a possibly-stale recorded name, and the slow path returns the *first* flow whose UUID matches, so colliding UUIDs cross-bind to the wrong flow:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/e83b94952901340de1e61e4aaaaf030e02672b30/griptape_nodes_library/engine/subflow_workflow_node.py#L232-L237

`after_node_deleted` uses this same resolver, so deleting one instance can delete another instance's child flow.

Reachable via (a) two Workflow nodes referencing the same child workflow when that child itself contains a Workflow node (3+ level nesting, exactly this PR's domain), or (b) duplicating / copy-pasting a `SubflowWorkflowNode` (node metadata round-trips through serialize/deserialize and `subflow_owner_uuid` is opaque to the engine, so it is not remapped).

**Impact:** Instances cross-bind to and can delete each other's child flows, re-creating the node<->flow clash this PR fixes, at higher multiplicity. This is not a regression of the single-instantiation case the PR targets.

**Coverage gap:** The integration test references `subflow_middle_echo` only once, and `test_two_nodes_same_workflow_bind_to_distinct_flows` uses distinct UUIDs (`U1`/`U2`). Nothing exercises two instances sharing a UUID.

**Suggested fix:** Make identity instance-scoped, for example by remapping each node/flow `subflow_owner_uuid` pair during import (and regenerating on node duplication), or by scoping the lookup to child flows of the node's containing flow and regenerating duplicate identities. Add coverage where two outer nodes import the same middle workflow that contains a nested Workflow node.

griptape_nodes_library/engine/subflow_workflow_node.py
```diff
@@ -76,21 +97,30 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
...
+        GriptapeNodes.handle_request(DeleteFlowRequest(flow_name=owned_subflow))
 
     def _on_refresh_workflow(self, button: Button, button_details: ButtonDetailsMessagePayload) -> None:  # noqa: ARG002
         workflow_name = self.get_parameter_value("workflow_file")
         if not workflow_name:
             return
         self._update_workflow_shape_parameters(workflow_name, preserve_matching=True)
         # Clear so _reload_subflow doesn't skip the reload for the same workflow.
-        self.metadata.pop("_subflow_workflow", None)
+        self.metadata.pop(SUBFLOW_WORKFLOW_KEY, None)
```
collindutter: > [!NOTE]

**Finding: refreshing an unclaimed legacy node leaks its existing child flow.**

`_on_refresh_workflow` pops `SUBFLOW_WORKFLOW_KEY` (this line) before calling `_reload_subflow`. For a modern, UUID-stamped node that is fine: `_find_owned_subflow_name` resolves the flow by UUID without needing the key, and bypassing the same-workflow early return then correctly deletes and re-imports the flow:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/e83b94952901340de1e61e4aaaaf030e02672b30/griptape_nodes_library/engine/subflow_workflow_node.py#L150-L163

But an *unclaimed legacy* node (pre-UUID child flow, never executed/adopted since load) has an unstamped flow, so UUID resolution misses it and it must fall back to the legacy resolver. That resolver requires `SUBFLOW_WORKFLOW_KEY`, which refresh has already popped:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/e83b94952901340de1e61e4aaaaf030e02672b30/griptape_nodes_library/engine/subflow_workflow_node.py#L253-L259

So `_recorded_subflow_if_unclaimed` returns `None`, `_reload_subflow` imports a fresh child flow, and the original legacy flow is never deleted.

**Impact:** The original legacy flow is orphaned. A later save or delete then preserves that orphan (the node now owns the freshly-imported flow), re-creating the teardown problems this PR addresses. Unlike the deliberate safety tradeoff in `after_node_deleted`, this leak is avoidable: refresh still holds both the recorded name and the workflow provenance before the pop.

**Coverage gap:** Unit tests cover the modern refresh path (`test_forces_reload_of_already_loaded_workflow`) but not a legacy refresh.

**Suggested fix:** Instead of popping `SUBFLOW_WORKFLOW_KEY` here, pass an explicit `force_reload` flag into `_reload_subflow` so it can resolve (using the intact key for legacy adoption) and delete the existing flow before bypassing its same-workflow early return. Add a legacy refresh test.

PR #480, opened 2026-07-23
fix(openai-image): upload reference images and send public URLs

zachgiordano:
Closes #474

## Problem

The `OpenAI Image Generation` node base64-inlined reference images into the JSON request body sent to the Griptape Cloud model proxy. Base64 encoding inflates payload size by ~33%, so a couple of large reference images could push the request body past the proxy's 100MB cap (`DATA_UPLOAD_MAX_MEMORY_SIZE`), causing a hard failure *before* generation even started.

## Fix

Upload each reference image to Griptape Cloud and pass the proxy a **public URL** instead of base64-inlined bytes. The URL keeps the request body tiny regardless of reference size.

- Verified empirically that the proxy forwards `images:[{image_url: ...}]` verbatim to OpenAI's `/v1/images/edits`, and that gpt-image accepts a plain HTTPS URL in `image_url`.
- Each reference is uploaded via a transient `PublicArtifactUrlParameter` (a hidden, PROPERTY-mode scratch parameter minted per upload). Already-public non-localhost `http(s)` URLs pass through unchanged and are not re-uploaded.
- Uploads are tracked and cleaned up (artifact deleted + scratch parameter removed) in `_process_generation`'s `finally` block, so nothing accumulates across runs — including when generation raises.
- The upload runs on the aprocess event loop (not a worker thread) because it resolves macro paths (e.g. `{inputs}/...`) via `GriptapeNodes.handle_request`, which needs the node's project context.

Follows the established multi-upload pattern from `seedance_2_0_video_generation.py`.

## Tests

- Reference image is uploaded and the payload carries the returned public URL.
- Already-public URLs pass through unchanged and mint no scratch parameter.
- Invalid/unreadable input raises a clear `Failed to prepare input image` error.
- Scratch parameters and uploaded artifacts are cleaned up after generation — parametrized to cover both the success path and the case where the underlying generation raises.

Clownin

### DIFF
--- a/griptape_nodes_library/image/openai_image_generation.py
+++ b/griptape_nodes_library/image/openai_image_generation.py
@@ -3,16 +3,20 @@ from __future__ import annotations
 import base64
 import logging
 import re
+from contextlib import suppress
 from typing import Any, ClassVar
+from uuid import uuid4
 
 from griptape.artifacts import ImageArtifact, ImageUrlArtifact
-from griptape_nodes.exe_types.core_types import ParameterGroup, ParameterList, ParameterMode
+from griptape_nodes.exe_types.core_types import Parameter, ParameterGroup, ParameterList, ParameterMode
+from griptape_nodes.exe_types.param_components.artifact_url.public_artifact_url_parameter import (
+    PublicArtifactUrlParameter,
+)
 from griptape_nodes.exe_types.param_components.project_file_parameter import ProjectFileParameter
 from griptape_nodes.exe_types.param_types.parameter_dict import ParameterDict
 from griptape_nodes.exe_types.param_types.parameter_image import ParameterImage
 from griptape_nodes.exe_types.param_types.parameter_int import ParameterInt
 from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
-from griptape_nodes.files.file import File, FileLoadError
 from griptape_nodes.traits.options import Options
 from griptape_nodes.utils.artifact_normalization import normalize_artifact_list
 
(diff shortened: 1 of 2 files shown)


griptape_nodes_library/image/openai_image_generation.py
```diff
@@ -492,6 +501,20 @@ def validate_before_node_run(self) -> list[Exception] | None:
...
+        try:
+            await super()._process_generation()
+        finally:
+            # Delete each uploaded reference and remove its scratch parameter. The scratch name is
+            # unique per upload, so leaving it would accumulate parameters on the node across runs.
+            for helper, scratch_name in self._pending_reference_uploads:
+                with suppress(Exception):
+                    helper.delete_uploaded_artifact()
+                with suppress(Exception):
+                    self.remove_parameter_element_by_name(scratch_name)
```
SavagePencil, line 515: is this how we do it on other nodes?

zachgiordano: Yeah, this pattern was introduced with [Seedance 2.0 Node](https://github.com/griptape-ai/griptape-nodes-library-standard/blob/main/griptape_nodes_library/video/seedance_2_0_video_generation.py#L563) I believe.

I think the "scratch parameter" technique here is compensating for the ParameterList with multiple inputs, like multiple ImageUrlArtifact(s). The PublicArtifactUrlParameter component we have currently doesn't play nicely with ParameterList. It can really only be used directly with a single input artifact. So we just borrow the PublicArtifactUrlParameter uploading functionality without messing with the ParameterList and how the Node makes use of those items.

Maybe it's worth cutting an engine issue to create a PublicArtifactUrlParameter component variant that is meant to handle lists? We could potentially abstract this ugliness out of the Node code then.

SavagePencil: yeah go for it but for now lemme approve to get these dudes unblocked

SavagePencil (approved):

PR #486, opened 2026-07-27
feat(list-files): support path-aware recursive glob patterns in match_pattern

shhlife:
## Summary

Closes #473

- `match_pattern` now supports path-aware patterns containing `/` or `**` (e.g. `**/outputs/images/*.png`), matched against the entry's path relative to `directory_path`
- Patterns without `/` or `**` (e.g. `*.png`, `report_*.txt`) are unchanged — still matched against basename via `fnmatch`
- Path patterns automatically enable recursion; no need to toggle the `recursive` checkbox
- Uses a small regex translator (`_compile_path_pattern`) so `**` correctly matches zero-or-more path components and `*` stays within a single segment — proper glob semantics, not `fnmatch`
- Both sides of `relative_to()` are resolved before comparison so symlinked roots (e.g. `/tmp` → `/private/tmp` on macOS) don't silently break matching
- `match_pattern_case_sensitive` is honoured in both modes
- `list_options`, `sort_by`, `sort_order`, `use_absolute_paths`, and `show_hidden` all still apply

Example where we can find all input images for every character:

image

## Test plan

- [ ] `match_pattern="**/outputs/images/*.png"` returns PNGs nested at any depth under an `outputs/images/` folder
- [ ] `match_pattern="*.png"` behavior is unchanged (basename match, no forced recursion)
- [ ] `match_pattern_case_sensitive=False` matches case-insensitively in both modes
- [ ] `list_options="List files only"` still filters correctly with a path pattern
- [ ] Empty `directory_path` with a path pattern returns a clear failure (not a silent empty list)
- [ ] Works on macOS where `/tmp` resolves via symlink to `/private/tmp`


griptape_nodes_library/files/list_files.py
```diff
@@ -214,6 +256,15 @@ def _collect_entries_recursive(
         collected: list = []
         stack: list[str | None] = [root_directory_path]
         visited: set[str] = set()
+        pattern = match_pattern.strip()
+        use_pattern = bool(pattern)
+        is_path_pattern = use_pattern and self._is_path_pattern(pattern)
+        root_resolved = str(Path(root_directory_path).resolve()) if root_directory_path and is_path_pattern else ""
```
cjkindel: AI Assisted Comment:

> **root_resolved is CWD-anchored but entry paths are workspace-anchored, so any non-absolute directory_path returns zero results with a SUCCESS status.**
> Path(root_directory_path).resolve() resolves against the process CWD and doesn't expand ~, env vars, or {project_dir} macros. The engine resolves the same string against the configured workspace ([file.py:196](vscode-webview://0c5bimdqi86ou9obd49pmqt5lqucfalvknrsanmog52ml4hisai8/nodes/griptape-nodes-engine/src/griptape_nodes/files/file.py#L196)), and entry.absolute_path derives from that. So with directory_path="assets" and match_pattern="**/outputs/images/*.png", relative_to raises for every entry, each falls back to its bare basename, the multi-segment pattern can't match a basename, and you get an empty list with was_successful=True. The GUI picker supplies absolute paths, which is exactly why this hides; **/*.png also still "works" via the basename fallback, making it look intermittent.

griptape_nodes_library/files/list_files.py
```diff
@@ -152,6 +155,45 @@ def __init__(
...
+            else:
+                result.append(re.escape(pattern[i]))
+                i += 1
+        return re.compile("".join(result), flags)
+
+    @staticmethod
+    def _match_path_entry(entry, root_resolved: str, compiled: re.Pattern) -> bool:
+        try:
+            # Both sides are resolved so symlinked roots (/tmp → /private/tmp on macOS) don't break relative_to.
+            candidate = Path(entry.absolute_path).resolve().relative_to(root_resolved).as_posix()
```
cjkindel: How worried are we about symlinks?

AI Assisted Comment
> **.resolve() on entry paths silently drops everything under a symlinked subdirectory.**
> The engine deliberately preserves symlinks ([os_manager.py:1644](vscode-webview://0c5bimdqi86ou9obd49pmqt5lqucfalvknrsanmog52ml4hisai8/nodes/griptape-nodes-engine/src/griptape_nodes/retained_mode/managers/os_manager.py#L1644)). With project/outputs -> /Volumes/bigdisk/outputs, the walk descends fine but every file resolves outside root_resolved, hits the basename fallback, and is excluded. Worth noting: flipping to .absolute() would break the /tmp -> /private/tmp case the comment on line 189 is guarding, so the fix is canonicalizing the root and not resolving entries.

griptape_nodes_library/files/list_files.py
```diff
@@ -152,6 +155,45 @@ def __init__(
...
+                result.append(".*")
+                i += 2
+            elif pattern[i] == "*":
+                result.append("[^/]*")
+                i += 1
+            elif pattern[i] == "?":
+                result.append("[^/]")
+                i += 1
+            else:
+                result.append(re.escape(pattern[i]))
```
cjkindel, line 202: AI Assisted Comment:

> **character classes silently stop working once a pattern gains a folder prefix.**
> img_[0-9].png matches img_3.png via fnmatchcase, but **/img_[0-9].png compiles to (?:.*/)?img_\[0\-9\]\.png and matches nothing. One field, two glob dialects chosen by the presence of a slash, with the tooltip documenting only * and **.

griptape_nodes_library/files/list_files.py
```diff
@@ -238,15 +289,20 @@ def _collect_entries_recursive(
...
-                collected.extend(
-                    self._filter_entries(
-                        [entry],
-                        include_files=include_files,
-                        include_folders=include_folders,
-                        match_pattern=match_pattern,
-                        match_pattern_case_sensitive=match_pattern_case_sensitive,
-                    )
-                )
+                include = (entry.is_dir and include_folders) or (not entry.is_dir and include_files)
```
cjkindel: AI Assisted Comment:

> **filter policy duplicated instead of generalizing _filter_entries.**
> The include/pattern logic is now inlined in the walk loop (with an assert at line 296 to appease the type checker) while _filter_entries keeps its own copy for the flat branch. That divergence is what forces the recursive = True workaround at line 331; pushing path-awareness into _filter_entries would remove both the duplication and the override.

griptape_nodes_library/files/list_files.py
```diff
@@ -152,6 +155,45 @@ def __init__(
             result_details_placeholder="Details on the list file attempt will be presented here.",
         )
 
+    @staticmethod
+    def _is_path_pattern(pattern: str) -> bool:
+        return "/" in pattern or "**" in pattern
```
cjkindel, line 160: AI Assisted Comment:

> **Windows-style outputs\*.png isn't recognized as a path pattern** 
> and falls through to basename matching, matching nothing silently. Weakest of the five (the tooltip teaches forward slashes), but a replace("\\", "/") before the check is cheap.

cjkindel (changes requested):

griptape_nodes_library/files/list_files.py
```diff
@@ -152,6 +155,45 @@ def __init__(
...
+            elif pattern[i] == "*":
+                result.append("[^/]*")
+                i += 1
+            elif pattern[i] == "?":
+                result.append("[^/]")
+                i += 1
+            else:
+                result.append(re.escape(pattern[i]))
+                i += 1
+        return re.compile("".join(result), flags)
```
collindutter, line 204: Are we working around the fact that `ListDirectoryRequest` only lists a single level? Should we add recursive functionality over there?

griptape_nodes_library/files/list_files.py
shhlife: Yep — if someone has project/outputs pointing to /Volumes/bigdisk/outputs, patterns like **/outputs/*.png should still find their files there.

griptape_nodes_library/files/list_files.py
shhlife: Good catch — done. _filter_entries now handles both `basename` and path patterns; the recursive branch delegates to it instead of inlining the logic. The recursive = True override stays because path patterns inherently need tree traversal, not because of the duplication.

griptape_nodes_library/files/list_files.py
shhlife: 👍 normalized at read time so the rest of the logic only sees forward slashes."

griptape_nodes_library/files/list_files.py
shhlife: @collindutter that's a great question.. if we imagine other nodes using this capability, we probably should add it in engine.

What do you think? Make a new request?

griptape_nodes_library/files/list_files.py
```diff
@@ -152,6 +155,65 @@ def __init__(
...
+                if j < len(pattern) and pattern[j] == "]":
+                    j += 1
+                while j < len(pattern) and pattern[j] != "]":
+                    j += 1
+                if j < len(pattern):
+                    char_class = pattern[i : j + 1]
+                    # Glob [!...] means "not"; regex spells that [^...].
+                    if char_class.startswith("[!"):
+                        char_class = "[^" + char_class[2:]
+                    result.append(char_class)
```
cjkindel, line 196: AI Assisted Comment:

> **malformed character range raises an uncaught re.PatternError and halts the flow.**
> The new branch forwards the class body verbatim into re.compile. **/[9-1].png — transposed digits, an easy typo — raises. I traced the severity: node_manager.py:3285 catches it, but node_executor.py:281 re-raises as RuntimeError, so the flow stops and the node's own failure control output is never taken. The user gets "bad character range z-a at position 9". The same pattern without a folder prefix returns empty harmlessly via fnmatchcase. A try/except re.error around the compile, using the existing _set_status_results pattern from line 346, closes it.

griptape_nodes_library/files/list_files.py
```diff
@@ -235,20 +306,25 @@ def _collect_entries_recursive(
...
-                        include_files=include_files,
-                        include_folders=include_folders,
-                        match_pattern=match_pattern,
-                        match_pattern_case_sensitive=match_pattern_case_sensitive,
-                    )
+            # Derive root from the engine's own resolution of directory_path — can't use
+            # Path.resolve() independently because the engine anchors relative paths to the
+            # workspace root, not the process CWD.
+            if is_path_pattern and not root_resolved and result.entries:
+                root_resolved = str(Path(result.entries[0].absolute_path).parent)
```
cjkindel: AI Assisted Comment:

> **root derivation is inferred from the first entry rather than read from the result.**
> The fix is directionally right but guessed at the value. if ... and result.entries never fires when the root has no visible entries (all hidden, or content only deeper), leaving root_resolved = "" → relative_to("") raises for every entry → zero matches with SUCCESS. And because the guard is not root_resolved rather than one-shot on the root pop, the first non-empty directory visited can be a subdirectory, which then becomes the root. The engine already returns the authoritative value as ListDirectoryResultSuccess.current_path (os_events.py:207, set at os_manager.py:1769) — using it deletes the inference.

shhlife: switched to result.current_path which the engine already authoritative-resolved. Fixes the empty-root case and removes the guessed inference from entries

cjkindel (approved):

PR #489, opened 2026-07-28
Fix: Trim Video accuracy

alex-rumsey-foundry:
## Summary
- Fixes #472: `Trim Video` in **frame range** mode could return more frames than requested (e.g. `0`–`81` overshot to 83 frames), while `0`–`41` worked correctly.
- Root cause (verified empirically, not the keyframe-snap theory in the issue): `build_video_segment_cmd`'s stream-copy fast path (`-c copy`) cuts on decode-order (DTS) packet boundaries rather than presentation order (PTS). On B-frame-encoded sources this pulls in extra trailing frames past the requested cutoff — reproduced directly with a synthetic B-frame H.264 clip (83 frames via stream copy vs. 81 via re-encode for the same `0`–`81` range).
- Added a `frame_accurate: bool = False` parameter to `build_video_segment_cmd` that forces the accurate re-encode path regardless of duration/start offset. `TrimVideo` and `SplitVideo` now pass `frame_accurate=True` only for their `"frame range"` mode; timecode-mode trims keep the existing stream-copy fast path (no behavior change there).
- Audited every other node using `-c:v copy` (`add_film_grain`, `flip_video`, `blur_video`, `hold_video_frames`, `add_rgb_shift`, `adjust_video_eq`, `add_vignette`, `resize_video`, `concatenate_videos`) — none combine it with a partial time/frame-range seek, so no other nodes are affected by this bug.

## Test plan
- [x] Reproduced the bug with a synthetic B-frame H.264 test video and confirmed `frame_accurate=True` fixes it (81 frames instead of 83) via direct `ffprobe -count_frames` check
- [x] Confirmed short ranges (previously correct via the re-encode path) still work
- [x] Confirmed timecode-mode trims on long, start-at-zero segments still take the stream-copy fast path (no perf regression)
- [x] `uv run ruff check` clean on changed files
- [x] `uv run pytest tests/unit` — 627 passed, 11 skipped

### DIFF
--- a/griptape_nodes_library/utils/ffmpeg_utils.py
+++ b/griptape_nodes_library/utils/ffmpeg_utils.py
@@ -436,6 +436,8 @@ def build_video_segment_cmd(
     start_sec: float,
     end_sec: float,
     output_path: str,
+    *,
+    frame_accurate: bool = False,
 ) -> list[str]:
     """Build an ffmpeg command to extract a video segment.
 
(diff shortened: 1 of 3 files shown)


shhlife (approved):
awesome!

PR #490, opened 2026-07-28
feat(seedance): add multimodal reference list input mode

### LINKED ISSUE #488: Can't wire a variable, mixed set of assets (images + audio) into Seedance 2.0 video-generation node & other multi-modal models
## Problem

A user wants to wire a **variable, unknown number of mixed assets** (images and audio) into a single Seedance video-generation node, where the exact set changes per shot. For example: one shot needs 3 image assets; another needs 1 image asset + 1 audio asset. They want a single Seedance node whose inputs adapt to whatever assets a given shot requires, rather than maintaining many IF/Then branches or many near-duplicate sub-workflows (which they correctly note is inefficient).

Today there's no clean way to do this:

- Connecting a list-producing node (e.g. `CreateImageList`) to the Seedance `Seedance20VideoGeneration` "Reference Images" `ParameterList` doesn't work — the connection can't be made. The same is reported for `OpenAiImageGeneration`'s "Input Images" parameter.
- Even if image-lists worked, the real need is an **arbitrary, mixed** list (images *and* audio), which no single list input currently models.

The desired outcome: a user can assemble an arbitrary, possibly-mixed set of assets via logic/list nodes and feed it into one Seedance node, with the node handling a dynamic number and mix of inputs — without per-shot branching or duplicated workflows.

## Origin

Reported by a user in the community Discord, then discussed by the team. Surface: Griptape Nodes (Seedance node + ParameterList inputs). Reproducible in both the UI and in code per the team discussion.

## Impact

Blocks a realistic multi-shot orchestration use case. The user's only current workarounds — IF/Then logic with many node variants, or many sub-workflows each hard-wired to a fixed asset layout — are inefficient and don't scale to an unknown asset count/mix.

## Repro / evidence

- Add a `CreateImageList` node and try to connect it to `Seedance20VideoGeneration`'s "Reference Images" `ParameterList` → cannot connect.
- Same with `OpenAiImageGeneration`'s "Input Images" parameter.

## Possible direction

Two complementary angles raised in discussion (captured as starting points, not decisions):

1. **General:** `ParameterList`s should accept a list as an input (connect a list-producing node directly to a list parameter). This pattern likely applies broadly to "list"-type inputs across many nodes, not just Seedance. Note: griptape-ai/griptape-nodes-engine#2779 ("Make sure ParameterList can take a list as one of the inputs") was closed as Done, but the behavior appears not to work in practice — worth checking whether this is a regression or an incomplete implementation. Related: griptape-ai/griptape-nodes-engine#473, griptape-ai/griptape-nodes-engine#4533, griptape-ai/griptape-nodes-engine#1450.
2. **Seedance-specific:** the node could accept a mixed audio/image list as an "input mode" and parse the asset types internally (e.g. accept a list of `*UrlArtifact` and route each by type).

Reported via Discord: https://discord.com/channels/1096466116672487547/1353916042778378333/1521088278458142810

feltech:
Closes #488.

Add support for an arbitrary list input mode to the Seedance 2.0 Video Generation node, "Multimodal Reference List", which is equivalent to "Multimodal References", but that takes a list as a single argument, where the list contains mixed audio, image and video references.

### DIFF
--- a/griptape_nodes_library/video/seedance_2_0_video_generation.py
+++ b/griptape_nodes_library/video/seedance_2_0_video_generation.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import json as _json
 import logging
+import os
 from contextlib import suppress
 from dataclasses import dataclass
 from typing import Any, ClassVar
@@ -40,6 +41,9 @@ from griptape_nodes_library.assets import (
 )
 from griptape_nodes_library.media import coerce_media_url_or_data_uri
 from griptape_nodes_library.proxy import GriptapeProxyNode
+from griptape_nodes_library.utils.audio_utils import is_audio_url_artifact
+from griptape_nodes_library.utils.image_utils import SUPPORTED_IMAGE_EXTENSIONS
+from griptape_nodes_library.utils.video_utils import SUPPORTED_VIDEO_EXTENSIONS, is_video_url_artifact
 
 logger = logging.getLogger("griptape_nodes")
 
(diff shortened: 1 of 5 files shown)


griptape_nodes_library/video/seedance_2_0_video_generation.py
```diff
@@ -350,6 +374,38 @@ def __init__(self, **kwargs: Any) -> None:
...
+                "Mixed list of reference images, audio, and videos, produced by a list node (e.g. Create List). Each"
+                f" item is routed by type (up to {MAX_REFERENCE_IMAGES} images, {MAX_REFERENCE_VIDEOS} videos,"
+                f" {MAX_REFERENCE_AUDIO} audio). Local videos are uploaded to a short-lived public URL on execution;"
+                " audio must be mp3 or wav format."
+            ),
+            allowed_modes={ParameterMode.INPUT},
+            ui_options={"display_name": "Reference Media"},
+        )
+        # Mirror the per-slot "Media Upload" badge shown by the individual
+        reference_media_param.set_badge(
```
feltech, line 396: [fix] Badge isn't showing in the UI

griptape_nodes_library/video/seedance_2_0_video_generation.py
```diff
@@ -837,89 +975,202 @@ async def _add_media_inputs_async(self, content_list: list[dict[str, Any]], para
...
+                        content_list,
+                        ref_image,
+                        idx,
+                        supports_assets=supports_assets,
+                        asset_label=f"reference_media image {idx}",
+                    )
+                    if tag:
+                        order_log_list.append(tag)
+
+                for idx, ref_video in enumerate(list_videos[:MAX_REFERENCE_VIDEOS], start=1):
```
feltech, line 1039: [question] Does video reference input definitely work (did a manual test and inconclusive if the output video used the input video).  Note that image and audio were definitely incorporated.

PR #494, opened 2026-07-29
Add SplitListByType

cjkindel:
Also includes types for lists so we can split them

https://www.loom.com/share/ad43e283b8cc40f392706a8b3d328c20

### DIFF
--- a/griptape_nodes_library/lists/create_bool_list.py
+++ b/griptape_nodes_library/lists/create_bool_list.py
@@ -11,7 +11,7 @@ class CreateBoolList(BaseCreateListNode):
             name,
             metadata,
             input_types=["bool"],
-            output_type="list",
+            output_type="list[bool]",
             default_value=None,
             items_tooltip="List of boolean items to add to",
             ui_options={"hide_property": False},
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -1897,6 +1897,22 @@
       }
     },
     {
+      "class_name": "SplitListByType",
+      "file_path": "griptape_nodes_library/lists/split_list_by_type.py",
+      "metadata": {
+        "category": "lists",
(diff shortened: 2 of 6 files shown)


griptape_nodes_library/lists/create_bool_list.py
```diff
@@ -11,7 +11,7 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
             name,
             metadata,
             input_types=["bool"],
-            output_type="list",
+            output_type="list[bool]",
```
SavagePencil, line 14: something in my lizard brain tells me this has some side-effect that I can't remember. Maybe @shhlife can recall? Is there a difference in handling generic `list` vs. `list[type]`?

SavagePencil: There may be some UI or convert-to-text behavior that this may impact

cjkindel: It should only expand the supported output types, I had Claude prove it via a local .py test file. Inputs are unaffected

> 
Target input_types | accepts list (before) | accepts list[bool] (after) | changed?
-- | -- | -- | --
["list"] | ✅ | ✅ | no
["list[any]"] | ✅ | ✅ | no
["any"] | ✅ | ✅ | no
["list[bool]"] | ❌ | ✅ | newly allowed
["list[str]"] | ❌ | ❌ | no
["bool"] | ❌ | ❌ | no

This is with https://github.com/griptape-ai/griptape-nodes-engine/pull/5211 as well, if you are worried about the release order we can update the library's required engine version =

SavagePencil: sweet. And I think `list` by itself doesn't expand in the UI, but `list[str]` will turn it into a widget with individual rows of string entries. So maybe it just makes it better when expanded!

SavagePencil (approved):

PR #497, opened 2026-07-30
refactor: extract ProviderSelectionComponent to eliminate provider-management duplication

ana-ureche:
Refactor provider selection to eliminate duplicate lines.

Testing of the changes:

https://github.com/user-attachments/assets/c10647d4-9174-4926-afc1-c58e0e5f1cb1

### DIFF
--- a/griptape_nodes_library/utils/provider_selection_component.py
+++ b/griptape_nodes_library/utils/provider_selection_component.py
@@ -17,12 +17,13 @@ _GRIPTAPE_CLOUD_PROVIDER = ProviderConfig(name="griptape_cloud", type="griptape_
 
 
 class ProviderSelectionComponent:
-    def __init__(self, node, model_param, *, gtc_model_choices, gtc_model_data):
+    def __init__(self, node, model_param, *, gtc_model_choices, gtc_model_data, default_model: str | None = None):
         # adds model_provider parameter to the node (buttons wired to self)
         self._node = node
         self._model_param = model_param
         self._gtc_model_choices = gtc_model_choices
         self._gtc_model_data = gtc_model_data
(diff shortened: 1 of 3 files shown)


griptape_nodes_library/utils/provider_selection_component.py
```diff
@@ -0,0 +1,142 @@
...
+        self, button: Button, button_details: ButtonDetailsMessagePayload
+    ) -> NodeMessageResult | None:  # noqa: ARG002
+        """Refresh the model dropdown for the currently selected provider."""
+        provider_name = self._node.get_parameter_value("model_provider") or "griptape_cloud"
+        self.update_model_choices_for_provider(provider_name)
+        return None
+
+    def update_model_choices_for_provider(self, provider_name: str) -> None:
+        if provider_name == "griptape_cloud":
+            # Use a curated vision-only subset rather than the full GTC model list.
```
cjkindel: Why are we intentionally truncating this to vision only? What if this is being used by Agent?

ana-ureche: You are right, I missed that part, I am on it now!

ana-ureche: The comment was a leftover from when I copied the code from `describe_image.py`, but I was passing down the model choices from the nodes so that was working right, just needed to update the name for it.
one bug that I caught was that the default for **agent** is not the first model in the list like it is for **describe image** so I just fixed that by giving default_model as a parameter to `ProviderSelectionComponent`

cjkindel (approved):

shhlife (approved):

collindutter (approved):
Thanks for taking the initiative on cleaning this up!

PR #507, opened 2026-08-04
feat(access): filter model dropdowns through OFFER_MODEL

collindutter:
Integrates most nodes that offer models with `ModelAccessComponent`. 

Tested by (automated):
1. Denying all models for my license.
2. Adding every node from this PR to the canvas.
3. Running every node.
4. Verifying denial error.

Part of #432.


griptape_nodes_library/utils/provider_selection_component.py
```diff
@@ -125,25 +136,36 @@ def _refresh_models_button(
...
             # Restore the component's per-row license decoration and badge; the
             # _update_option_choices call above only refreshed choices and value.
             self._model_access.reinstall_options()
             return
+        # A third-party provider's models are not catalog models -- offer whatever the
+        # provider itself reports, never `self._gtc_model_choices` (those are catalog keys,
+        # meaningless and ungate-able for this provider's own driver).
         models = self.fetch_models_for_provider(provider_name)
-        default = models[0] if models else self._default_model
+        default = models[0] if models else ""
```
cjkindel: AI Assisted Comment:
> *Empty provider model list wedges the model dropdown*
> Switching to a third-party provider whose model fetch fails or returns empty now passes `choices=[]` / `default=""` to `_update_option_choices`. The engine sets the Options trait's choices to `[]` *before* raising `ValueError("Default model '' is not in the provided choices.")`, and nothing up-stack catches it (Agent/DescribeImage `after_value_set`, the refresh-models button). The set request fails and the model dropdown is left with zero choices — a state that persists through save/reload via the serialized `simple_dropdown` — so every later model assignment hits `IndexError` in the Options converter (`self.choices[0]` on `[]`).
>
> Repro: set `model_provider` to a configured provider that's unreachable (e.g. Ollama not running) or whose api_key secret is unset; `fetch_models_for_provider` swallows the error and returns `[]`. Pre-PR this path fell back to the non-empty GTC list, so the empty-choices case is new. The third-party branch needs to handle `[]` explicitly (skip the update and surface an error) rather than emptying the trait.

griptape_nodes_library/agents/agent.py
```diff
@@ -170,6 +227,7 @@ def strip_whitespace(value: str) -> str:
             parameter=model_param,
             model_choices=MODEL_CHOICES,
             default_model=DEFAULT_MODEL,
+            deprecated_values=LEGACY_MODEL_VALUES,
```
cjkindel: AI Assisted Comment:
> *Legacy-value converter rewrites third-party model selections*
> The legacy-value migration converter is attached to the shared `model` parameter permanently — `add_converter` is never undone, and `_update_option_choices` only mutates the trait's choices — so it keeps running after the provider switches away from Griptape Cloud. `_convert_legacy_value` is an unconditional dict lookup, and `LEGACY_MODEL_VALUES` contains bare ids that real third-party endpoints serve (`o1-mini`, `gemini-2.0-flash`, `claude-sonnet-4-6`, ...).
>
> On an OpenAI-compatible provider that lists both generations (Google's compat endpoint serves `gemini-2.0-flash` and `gemini-2.5-flash`), picking `gemini-2.0-flash` gets silently rewritten and the run executes on a different model than the user chose; if the replacement *isn't* in the provider's list, the Options validator rejects a legitimate selection with "Choice not allowed". Fires on interactive sets and on workflow load. Same exposure in DescribeImage.
>
> The migration should probably be gated on the active provider being griptape_cloud, or only rewrite when the current value isn't among the offered choices.

griptape_nodes_library/agents/agent.py
```diff
@@ -72,6 +70,81 @@
 SERVICE = "Griptape"
 DEFAULT_MODEL = "claude-sonnet-5"
 
+# Migrates values saved before the dropdown stored the provider's own model id. Folds in
+# both display labels once shown in the dropdown, catalog keys saved during the interval
+# the dropdown stored those instead, and this node's own former DEPRECATED_MODELS dict
+# (provider ids for models Griptape Cloud has fully retired, mapped to their replacement).
+LEGACY_MODEL_VALUES = {
```
cjkindel: AI Assisted Comment:
> *68-entry legacy map duplicated across three nodes*
> This 68-entry dict is byte-identical to the one in `config/prompt/griptape_cloud_prompt.py`, and `describe_image.py` carries a 32-entry exact subset. Can the cloud legacy-value map be defined once (next to `MODEL_CHOICES`) and imported/subset? The next retirement has to land in every copy in lockstep or the same saved workflow migrates differently per node — and the per-node migration test only checks each node's own table, so a divergence passes green.

griptape_nodes_library/text/summarize_text_task.py
```diff
@@ -9,9 +9,26 @@
...
+MODEL_CHOICES = [
+    "gpt-4.1",
+    "gpt-4.1-mini",
+    "gpt-4.1-nano",
+    "gpt-5",
+]
 DEFAULT_MODEL = "gpt-4.1-nano"
 
+# Migrates values saved before the dropdown stored the provider's own model id.
+LEGACY_MODEL_VALUES = {
```
cjkindel: AI Assisted Comment:
> *Six BaseTask nodes duplicate the model map and gating wiring*
> All six BaseTask nodes carry AST-identical `LEGACY_MODEL_VALUES`/`MODEL_CHOICES`, and each hand-wires the same trio — component construction, `after_value_set` → `on_value_set` forwarding, `raise_if_selection_denied()` at the top of `process()` — that `BaseDriver._install_model_access` centralizes for the driver family. Could BaseTask own both the constants and the wiring? As-is, the next task node has three call sites to remember with no enforcement: forget the forwarding and legacy values never migrate; forget the `process()` gate and a denied model executes.

griptape_nodes_library/proxy/griptape_proxy_node.py
```diff
@@ -80,6 +84,13 @@ def __init__(self, **kwargs: Any) -> None:
         self._api_key_provider: ProxyAuthProviderParameter | None = None
         self._initialize_api_key_provider()
 
+        # Assigned by subclasses whose model selection is a license-filtered dropdown:
+        # they construct a `ModelAccessComponent` over their model parameter and store it
+        # here, which is what wires the dropdown into `after_value_set`,
+        # `_get_api_model_id`, `_get_catalog_model_id`, and the `_submit_and_poll` gate.
+        # Stays None on the subclasses bound to a single model.
+        self._model_access: ModelAccessComponent | None = None
```
cjkindel, line 92: AI Assisted Comment:
> *36 proxy subclasses hand-wire ModelAccessComponent*
> 36 subclasses construct `ModelAccessComponent` inline (same copy-pasted "License-policy dropdown:" comment) while the base only declares this slot. Since every `_model_access is None` fallback silently degrades to ungated raw-parameter behavior — a new proxy node written the pre-PR way ships with no dropdown gating, no badges, no pre-upload gate, and nothing fails loudly — it'd be worth giving this base the same `_install_model_access` helper `BaseDriver` got (which also performs the default-in-choices check the inline sites skip).

griptape_nodes_library/config/base_driver.py
```diff
@@ -110,6 +116,99 @@ def params_to_sparse_dict(
...
+        self.set_parameter_value(param, default_model)
+        self._model_access = ModelAccessComponent(
+            node=self,
+            parameter=parameter,
+            model_choices=model_choices,
+            default_model=default_model,
+            deprecated_values=deprecated_values,
+        )
+
+    def _get_selected_model_id(self) -> str:
```
cjkindel, line 161: AI Assisted Comment:
> *Selected-model accessor and forwarding duplicated three ways*
> Body is identical to `GriptapeProxyNode._get_selected_model_id` (the docstrings are already drifting), and the `after_value_set` → `on_value_set` forwarding is now copied here, in the proxy base, and a third time in `agent.py`. A `selected_model_id` accessor on `ModelAccessComponent` itself (or a small shared mixin) would collapse all copies — otherwise a later fix to how the selected value is read lands in one hierarchy and not the other.

griptape_nodes_library/utils/provider_selection_component.py
```diff
@@ -125,25 +136,36 @@ def _refresh_models_button(
...
         if provider_name == "griptape_cloud":
+            # The component's own choices ARE catalog model keys; offer them as-is.
             default = self._model_access.pick_permitted_default() or self._default_model
             self._node._update_option_choices(param="model", choices=self._model_access.model_choices, default=default)
             # Restore the component's per-row license decoration and badge; the
             # _update_option_choices call above only refreshed choices and value.
             self._model_access.reinstall_options()
             return
+        # A third-party provider's models are not catalog models -- offer whatever the
+        # provider itself reports, never `self._gtc_model_choices` (those are catalog keys,
```
cjkindel: AI Assisted Comment:
> *_gtc_model_choices is now dead state*
> `self._gtc_model_choices` is write-only after this PR — the three fallback reads are gone and it's only referenced in this comment and a docstring. The ctor arg (and the `_default_model` fallback it feeds at line 53) also duplicates `model_access.model_choices`; both call sites pass the same list to both. Worth dropping the field and parameter.

griptape_nodes_library/config/base_driver.py
```diff
@@ -110,6 +116,99 @@ def params_to_sparse_dict(
...
+        """The license denial for the stored model selection, as a validation failure.
+
+        Returns `None` when the node never installed a license-filtered dropdown,
+        or when its current selection is permitted. Routes through the same
+        `raise_if_selection_denied` the execute path uses, so a denied model reads
+        identically whether it surfaces during validation or at execution.
+        """
+        if self._model_access is None:
+            return None
+        try:
```
cjkindel: AI Assisted Comment:
> *Exception-as-control-flow swallows unrelated RuntimeErrors*
> This calls `raise_if_selection_denied` and immediately catches its own `RuntimeError` to build the return list — and the except also swallows any *unrelated* RuntimeError from inside the component, reporting it as a license denial instead of surfacing it as a bug. The proxy path already uses the query form: `denial = self._model_access.selection_denial()` then `return [RuntimeError(...)] if denial else None`.

tests/unit/test_legacy_model_value_migration.py
```diff
@@ -0,0 +1,260 @@
...
+
+LIBRARY_NAME = "Griptape Nodes Library"
+
+# Every node type that installs a model-access dropdown with a `deprecated_values`
+# table, discovered by constructing each node type declared in the library and
+# checking for one. `TopazImageEnhance` is deliberately excluded: its `operation`
+# dropdown is a plain switch (denoise / enhance / sharpen), not a license-gated
+# model choice, so it never installs `ModelAccessComponent` (see
+# `tests/unit/test_proxy_model_access.py::test_topaz_image_enhance_operation_has_no_model_access`).
+NODE_TYPES_WITH_MODEL_ACCESS: list[str] = [
```
cjkindel, line 43: AI Assisted Comment:
> *Test coverage keyed off hand-maintained node lists*
> Coverage here is keyed off this hand-maintained 54-entry list (plus the 36-entry `NODE_MODEL_CASES` in `test_proxy_model_access.py`), and `_create_node` is duplicated verbatim in 8 test files — `test_model_dropdown_vocabulary.py` already imports private helpers from a sibling test module to avoid a 9th copy. The next node that installs a model dropdown is invisible to all of these suites unless its author appends it to two static lists in unfamiliar files. Could discovery iterate the library's declared node types, with `_create_node` moved to `tests/unit/conftest.py`?

griptape_nodes_library/agents/agent.py
collindutter: > [!NOTE]
> Beep boop. AI-assisted reply. Not my own words, but I reviewed it before posting.

Confirmed, and it predates this branch. `main`'s `Agent.before_value_set` already rewrites `DEPRECATED_MODELS` with no provider check, and that dict already contains the ambiguous bare ids you named: `gemini-2.0-flash`, `o1-mini`, `claude-sonnet-4-6`, `claude-3-7-sonnet`. So the wrong-model outcome is inherited rather than introduced. What this branch does change is the reach: the rewrite moved from `before_value_set`, which workflow load skips (`skip_before_value_set=True`), to a converter, which load runs. A saved third-party workflow now gets rewritten on load where it previously would not have been.

There is no library-only fix. The converter is installed inside `ModelAccessComponent`, `Parameter` exposes `add_converter` with no removal counterpart, and `_converters` is private. Nor can the node simply stop passing `deprecated_values`: that table is also what puts the legacy keys into `Options.choices`, and the `Options` trait converter runs *before* directly-attached ones, so without them a stored legacy value is rewritten to `choices[0]` before the migration ever sees it. Dropping the table trades a third-party bug for a Griptape Cloud one that hits every saved workflow.

Two related defects turned up while testing this, both on engine `main`:

- `ModelAccessComponent.reinstall_options()` appends a second `Options` trait rather than replacing the first. Its docstring claims `add_trait` replaces; `add_trait` calls `add_child`.
- `BaseNode._update_option_choices` writes a manual `ui_options["simple_dropdown"]`, and `Parameter.ui_options` merges trait options *underneath* the manual ones and returns a fresh dict, so a reinstalled trait can never displace it. Since that key is what `Options.choices` actually reads, the legacy keys never come back after a provider round trip: on `Agent`, `"GPT-4o"` becomes `claude-sonnet-5` instead of `gpt-4o`.

Fixes:

- griptape-ai/griptape-nodes-engine#5286 adds `Parameter.remove_converter` and `ModelAccessComponent.legacy_value_converter`, and fixes `reinstall_options`.
- #531 is the library side: detach the component's converter, reinstall the same migration behind a provider check via the already-public `migrate_value`. Draft until that engine change releases and the pin moves.
- griptape-ai/griptape-nodes-engine#5288 tracks the root cause, which is that a one-time load-time data migration is implemented as a permanent value coercion running on every assignment from every source. If that lands, the provider check and the library commit both get deleted rather than merged.

Also worth flagging on this thread: 8a6f69f dropped Agent's `model_deprecation_notice` with no rationale in the commit message, so migrations that `main` announced ("the model was updated to X, please save your workflow") are now silent. Happy to restore it here or take it as a follow-up.

griptape_nodes_library/proxy/griptape_proxy_node.py
collindutter: > [!NOTE]
> Beep boop. AI-assisted reply. Not my own words, but I reviewed it before posting.

Leaving construction inline here, for two reasons.

The concrete argument for the helper no longer holds: the default-in-choices check is now performed by the engine component itself, in `ModelAccessComponent._raise_if_misconfigured`, which raises when `default_model` is not among `model_choices`. `BaseDriver._install_model_access` duplicates that check rather than adding it. So a proxy helper would once again forward every argument unchanged, which is exactly what 53a31c1 removed.

More importantly, a helper does not close the gap you are actually pointing at. Forgetting `self._install_model_access(...)` is precisely as silent as forgetting `ModelAccessComponent(...)`: either way `_model_access` stays `None`, and every read in this base class is written to fall back to raw-parameter behaviour. What would close it is a test that walks every node type whose manifest declares gated models and asserts the node installs a component, so the omission fails loudly instead of shipping ungated. That is the ground your other comment on the test suites covers, and it is deferred this pass.

Where base-class ownership did buy something beyond ergonomics, it is in this branch: `BaseTask` now owns the shared model list, the migration table, `_add_model_parameter()`, the `after_value_set` forwarding, and `_require_permitted_model()`, which gates before returning the selection. That last one is the part worth copying if we revisit the proxy family, because it removes the ungated read rather than renaming the wiring: a task node cannot read its dropdown without passing the license gate.

cjkindel (approved):

PR #499, opened 2026-07-31
feat: add Seedance 2.5 to the Seedance video generation node

collindutter:
Requires https://github.com/griptape-ai/griptape-cloud/pull/2144 to test.

PR #520, opened 2026-08-07
feat: add a dedicated Seedance 2.5 video generation node

zachgiordano:
## Why a new node

Seedance 2.5 (`dreamina-seedance-2-5-260628`) is not "2.0 with bigger numbers". It classifies every request into one of five task types from `content.role` **plus prompt intent**, and applies per-task constraints 2.0 doesn't have:

| Task type | `ratio` | `duration` | Extra requirement |
|---|---|---|---|
| Text-to-video | any | `[4,30]` or `-1` | — |
| First-frame / first-last-frame | **`adaptive` only** | `[4,30]` or `-1` | `role` = `first_frame` / `last_frame` |
| Reference-to-video | any | `[4,30]` or `-1` | Avoid edit/extend wording or it misclassifies |
| Video editing | **`adaptive` only** | **`-1` only** | Prompt must contain an edit keyword; input video 4–30s |
| Video extension | **`adaptive` only** | `[4,30]` or `-1` | Prompt must contain an extend keyword |

There is **no `task_type` field** — violations surface as an **asynchronous** `InvalidParameter.TaskTypeConstraint` error only after the task has queued and started processing. A node that can't distinguish these five cases can't fail fast, so users burn queue time on avoidable errors.

Add to that: 2.5-only `output_format` (mp4/mov), audio-only input (2.0 forbids it), no 1080p/4k, no `seed`/`camera_fixed`/`frames`, 30s duration, 50 reference assets. Folding all of that into the 2.0 node's three-mode selector means every tooltip, choice list, and validation branch becomes conditional on the model.

Closes #499 and #518 in favour of this approach. Both were good references — #518's `reference_videos` ParameterList rework is carried over here.

## What's in it

**`Seedance25VideoGeneration`** — a five-value `task` selector backed by a frozen `TaskConstraints` table that is the single source of truth for parameter visibility, `ratio`/`duration` narrowing, and validation. `ratio` and `duration` are `Options` (not sliders) specifically so their choice lists can be narrowed per task and an out-of-range current value coerced.

Reference inputs are three `ParameterList`s (30 images / 10 videos / 10 audio); list order is what `@Image N`, `@Video N`, and `@Audio N` resolve against in the prompt. Editing and extension validate for a trigger keyword before submission, since that wording is what the provider classifies on. Extras: `output_format` (mp4/mov, rewrites the output filename), `watermark`, `return_last_frame` (second download into `last_frame_file`; a failure there logs rather than failing the node, since the video already succeeded and was billed).

**`video/seedance_common.py`** — the private-asset (`asset://`) registration flow and media-URL helpers, behind `SeedanceProxyNode`, which both Seedance nodes now derive from. `_resolve_public_url_for_media` is taken in its #518 form (on-demand scratch upload rather than per-slot parameters), because a ParameterList of videos has no fixed slots to hang helpers off. The 2.0 node loses 403 lines and keeps its parameters, model list, capability table, and messages unchanged.

Registration: `gtc_seedance_2_5` catalog entry + `model_usage` declaration, and `Seedance25VideoGeneration: SEED` in the provider map. No `library_version` bump — that's the CI-generated PR's job.

## Two deviations from the plan, both for cause

**The 50-asset total check is unreachable.** The per-kind caps are 30 + 10 + 10 = exactly 50, so any input passing the per-kind checks already satisfies the total. I dropped the branch and replaced its test with one pinning the arithmetic, so bumping any single cap fails loudly instead of silently exceeding the total.

**Two conformance suites were silently losing coverage.** `test_griptape_proxy_node_byok.py` and `test_griptape_proxy_node_polling.py` match `GriptapeProxyNode` as a direct AST base name. Routing the 2.0 node through an intermediate base made them try to instantiate the abstract `SeedanceProxyNode` and, more importantly, **dropped `Seedance20VideoGeneration` from their coverage entirely**. Both now resolve base names to a fixed point and skip abstract classes; all three concrete Seedance nodes are collected. This is a test-infrastructure edit outside the original scope, but leaving it would mean a node quietly falling out of a library-wide gate.

## Testing

`make check` clean; `make test/unit` → **724 passed, 11 skipped**. 60 new unit tests for the 2.5 node; the 2.0 node's 44 tests pass with a 2-line import change.

Two integration workflows added and documented in `SEEDANCE_TESTING.md`: `test_seedance_2_5.py` (text-to-video) and `test_seedance_2_5_first_frame.py` (First/Last Frame with `output_format=mov`, exercising the adaptive-ratio lock).

**Draft pending end-to-end testing.** Three things need confirming against a live proxy:

- [x] **Private assets on 2.5.** `SUPPORTS_PRIVATE_ASSETS = True` follows #499's assumption, which noted it needs griptape-cloud #2144 to test. If the backend doesn't link provider assets for `dreamina-seedance-2-5-260628`, that flag needs flipping and the `SeedanceHumanReferenceAsset` wording adjusting.
- [x] **Trigger-keyword strictness.** BytePlus documents examples ("such as…"), not an exhaustive list, so a client-side check can reject a prompt the provider would have accepted. The generous keyword list plus the prompt badge is the mitigation; if it proves too strict, downgrade the editing/extension check from an error to a warning badge.
- [x] **The task cases I couldn't script** without a source video, tracked as TODOs in `SEEDANCE_TESTING.md`: Video Editing, Video Extension, `return_last_frame` wiring, reference-to-video with one image + one video, and a deliberate `ratio=16:9`-on-editing violation to confirm the client-side error fires before submission.

Seedance_2_5_Video_editing

Seedance_2_5_HumanAssetFlow

### DIFF
--- a/griptape_nodes_library.json
+++ b/griptape_nodes_library.json
@@ -406,6 +406,12 @@
                 "provider_model_id": "dreamina-seedance-2-0-mini-260615",
                 "key_support": "SUPPORTS_CUSTOMER_KEY_OR_GRIPTAPE_KEY"
               },
+              "gtc_seedance_2_5": {
+                "display_name": "Seedance 2.5",
+                "family": "Seedance",
+                "provider_model_id": "dreamina-seedance-2-5-260628",
+                "key_support": "SUPPORTS_CUSTOMER_KEY_OR_GRIPTAPE_KEY"
+              },
               "gtc_omnihuman_1_0": {
                 "display_name": "OmniHuman 1.0",
                 "family": "OmniHuman",
@@ -2437,11 +2443,37 @@
       }
     },
     {
+      "class_name": "Seedance25VideoGeneration",
+      "file_path": "griptape_nodes_library/video/seedance_2_5_video_generation.py",
+      "metadata": {
+        "category": "video",
+        "description": "Generate a video using Seedance 2.5 with multimodal references, editing, and extension via Griptape Cloud model proxy.",
+        "display_name": "Seedance 2.5 Video Generation",
+        "icon": "Sparkles",
+        "group": "create",
+        "tags": [
+          "video",
+          "generation",
+          "ai",
+          "api",
+          "seedance"
+        ],
+        "declarations": [
+          {
+            "type": "model_usage",
+            "model_ids": [
+              "gtc_seedance_2_5"
+            ]
+          }
+        ]
+      }
+    },
+    {
       "class_name": "SeedanceHumanReferenceAsset",
       "file_path": "griptape_nodes_library/video/seedance_human_reference_asset.py",
       "metadata": {
         "category": "video/seedance",
-        "description": "Package an image, video, or audio as a private-asset reference for Seedance 2.0 human-reference inputs.",
+        "description": "Package an image, video, or audio as a private-asset reference for Seedance human-reference inputs.",
         "display_name": "Seedance Human Reference Asset",
         "icon": "Link",
         "group": "create",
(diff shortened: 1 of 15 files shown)


cjkindel (approved):
Let's get eyes on it w/ latest before we cut a release

PR #526, opened 2026-08-11
Fix ResolveMacroPath nodes returning relative path instead of absolute

shhlife:
## Summary

- `GetPathForMacroResultSuccess` exposes both `resolved_path` (relative) and `absolute_path` (absolute); `BaseResolveMacroPath._resolve_one` was reading the wrong field
- One-line fix: `result.resolved_path` → `result.absolute_path` in the shared base helper, which fixes both `ResolveMacroPath` and `ResolveMacroPaths`
- Every other consumer in the engine (`files/file.py`, `files/directory.py`, `static_files_manager`, etc.) already uses `absolute_path` — these two nodes were the outliers

Closes #525.

## Test plan

- [ ] Run a `Resolve Macro Path` node with a macro like `{inputs}/my_file.png` and confirm the output is now a full filesystem path (e.g. `/Users/me/project/inputs/my_file.png`)
- [ ] Same for `Resolve Macro Path List` with multiple inputs
- [ ] Confirm a downstream node consuming the output can open the file without path resolution issues


mcaloud (commented):
What do you think of adding a unit test for this @shhlife? I see various tests in the engine code but none in here

mcaloud (approved):
Looks good!

PR #533, opened 2026-08-14
Fix/ffmpeg logs

cjkindel:
quiet means we get no error logs


shhlife (changes requested):
This looks like it's got some seedream image gen updates in it?  are they supposed to be in this pr?

shhlife (approved):

PR #534, opened 2026-08-15
feat: add aspect ratio resize mode to RescaleImage node

### LINKED ISSUE #295: Add 'aspect ratio' resize_mode to Rescale Image node
## Problem

The Rescale Image node (`griptape_nodes_library/image/rescale_image.py`, `RescaleImage`) supports `resize_mode` values of `width`, `height`, `percentage`, and `width and height`. Add a new `aspect ratio` resize mode that lets the user specify a target aspect ratio as a string and resizes the image to match it.

### Desired behavior

Add a new resize mode (e.g. `aspect ratio`) with a string parameter that accepts the ratio in **either** of two formats:

- **Ratio notation** — e.g. `16:9`. When the input contains a `:`, parse the two sides, compute the scalar (`16 / 9 = 1.77…`), and resize so the output matches that width:height ratio.
- **Scalar** — e.g. `1.77`. When the input is a plain number, use it directly as the width-to-height ratio.

The resulting dimensions should be derived from the chosen aspect ratio while honoring the existing `fit_mode` (`fit` / `fill` / `stretch`) so the new mode reuses the established letterbox / crop / stretch behavior in `_resize_with_fit_mode` rather than introducing a parallel path.

### UI

- **Placeholder text:** `16:9`
- **Tooltip:** explain that the user can provide the aspect ratio in either format — ratio notation like `16:9` or a scalar like `1.77`.
- Wire visibility into `after_value_set` the same way the other modes show/hide their relevant parameters (and surface `fit_mode` for this mode, since it's used).

### Validation / errors

If the input string matches neither format (not parseable as `N:M` and not parseable as a float), raise a clear error telling the user to provide the aspect ratio as either a ratio (e.g. `16:9`) or a scalar (e.g. `1.77`). Add this to `_validate_custom_parameters` alongside the existing per-mode validation.

## Origin

Requested by a team member during an ops triage session. There is an existing `aspect_ratio.py` and `get_closest_aspect_ratio.py` in the same image library that may be worth referencing for consistent parsing/handling.

## Impact

Quality-of-life enhancement — lets users resize to a known aspect ratio without manually computing target width/height pixel values.

shhlife:
## Summary

- Adds a new `aspect ratio` resize mode to the `RescaleImage` node
- Accepts the target ratio as ratio notation (`16:9`) or a decimal scalar (`1.77`)
- Reuses the existing `fit_mode` (fit/fill/stretch) via `_resize_with_fit_mode` — no parallel resize path
- Adds a scan button on the `aspect_ratio` parameter that auto-detects the ratio from the connected input image using GCD reduction

Closes #295

## Test plan

- [ ] Switch `resize_mode` to `aspect ratio` — verify `aspect_ratio` and `fit_mode` params appear, others hide
- [ ] Enter `16:9` → landscape image resizes to 16:9 ratio
- [ ] Enter `1.77` → same result as above (scalar path)
- [ ] Enter `9:16` → portrait crop/letterbox/stretch behaves correctly per fit mode
- [ ] Each fit mode (fit/fill/stretch) produces expected output
- [ ] Click the scan button with an input image connected → `aspect_ratio` field populates with GCD-reduced ratio
- [ ] Click the scan button with no input image → graceful failure message
- [ ] Enter garbage string (e.g. `foo`) → validation error surfaced before run

### DIFF
--- a/griptape_nodes_library/image/rescale_image.py
+++ b/griptape_nodes_library/image/rescale_image.py
@@ -1,8 +1,10 @@
+from math import gcd
 from typing import Any
 
-from griptape_nodes.exe_types.core_types import Parameter, ParameterGroup
+from griptape_nodes.exe_types.core_types import NodeMessageResult, Parameter, ParameterGroup
 from griptape_nodes.exe_types.param_types.parameter_int import ParameterInt
 from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
+from griptape_nodes.traits.button import Button, ButtonDetailsMessagePayload
 from griptape_nodes.traits.color_picker import ColorPicker
 from griptape_nodes.traits.options import Options
 from griptape_nodes.traits.slider import Slider
@@ -19,6 +21,7 @@ class RescaleImage(BaseImageProcessor):
     RESIZE_MODE_HEIGHT = "height"
     RESIZE_MODE_PERCENTAGE = "percentage"
     RESIZE_MODE_WIDTH_HEIGHT = "width and height"
+    RESIZE_MODE_ASPECT_RATIO = "aspect ratio"
 
     # Target size constants (for width/height modes)
     MIN_TARGET_SIZE = 1
@@ -46,6 +49,7 @@ class RescaleImage(BaseImageProcessor):
                 self.hide_parameter_by_name("target_size")
                 self.hide_parameter_by_name("target_width")
                 self.hide_parameter_by_name("target_height")
+                self.hide_parameter_by_name("aspect_ratio")
                 self.hide_parameter_by_name("fit_mode")
                 self.hide_parameter_by_name("background_color")
             elif value == self.RESIZE_MODE_WIDTH_HEIGHT:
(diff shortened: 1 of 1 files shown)


griptape_nodes_library/image/rescale_image.py
```diff
@@ -127,7 +143,27 @@ def _setup_custom_parameters(self) -> None:
...
+            ParameterString(
+                name="aspect_ratio",
+                default_value="16:9",
+                placeholder_text="16:9",
+                tooltip=(
+                    "Target aspect ratio. Use ratio notation (e.g. 16:9) or a decimal scalar (e.g. 1.77). "
+                    "The image is resized to match this ratio using the selected fit mode."
+                ),
+                traits={
+                    Button(
```
hutchinson, line 156: [question] does an `Options()` trait make sense here with common aspect ratio presets?

hutchinson (approved):
Looks good to go and a very useful addition! Just one question but not blocking a merge!

griptape_nodes_library/image/rescale_image.py
shhlife: @hutchinson I thought about that, but wanted to make sure people could plug into it.. we have a bigger task somewhere about creating a clear list of aspect ratio dropdowns? we can do this as a followup!

PR #536, opened 2026-08-17
feat: add VideoCapture node with webcam recording widget

### LINKED ISSUE #1243: Record Movie Webcam Node
We have a webcam node that lets you take a snapshot from your camera. It'd be great to also have a node that lets you record a movie from the webcam.

shhlife:
## Summary

https://github.com/user-attachments/assets/14e0b7cd-6f3a-4bf1-9b2b-63e3622a460b

- Adds a `VideoCapture` node that captures video and audio from the user's webcam and outputs a `VideoUrlArtifact`
- Custom JS widget handles live preview, recording with audio, playback review, and accept/re-record flow
- Accepts video via HTTP PUT to the static file server (raw bytes, no base64) — avoids WebSocket size limits for longer recordings
- Output path managed by `ProjectFileParameter`, respecting the project's `SAVE_NODE_OUTPUT` situation

## How it works

1. User records → clicks Accept
2. Widget requests an upload URL from Python via the WebSocket (`state: "requesting_upload_url"`)
3. Python allocates the indexed file path and returns a signed PUT URL + portable artifact path
4. Widget PUTs the raw MP4 blob directly to the static file server
5. Widget sends the artifact URL back to Python; Python emits the `VideoUrlArtifact` output

Closes https://github.com/griptape-ai/griptape-nodes-engine/issues/1243

## Test plan

- [ ] Add a VideoCapture node to a workflow
- [ ] Grant camera/mic permissions when prompted
- [ ] Record a clip, stop, review playback
- [ ] Click Accept — encoding overlay appears, then hides when processing completes
- [ ] `output_video` port emits a `VideoUrlArtifact` pointing to the saved `.mp4`
- [ ] Re-record (discard) and accept again — new indexed file created, output updates
- [ ] Change `output_file` parameter to a custom filename — saved to that name
- [ ] Run the node in a workflow — output propagates downstream correctly

### DIFF
--- a/griptape_nodes_library/video/video_capture.py
+++ b/griptape_nodes_library/video/video_capture.py
@@ -45,11 +45,12 @@ class VideoCapture(DataNode):
                 artifact = self._save_recording(value)
                 self.parameter_output_values["output_video"] = artifact
                 self.publish_update_to_parameter("output_video", artifact)
-                # Replace stored recording with a slim reference so the workflow
-                # never serialises the base64 blob, and signal the JS overlay to hide.
+                # Replace with a slim reference immediately — set_parameter_value triggers
+                # its own after_value_set → super() chain for the processed dict, so we
+                # return here to prevent the blob-carrying value from ever reaching super().
                 processed = {"state": "processed", "url": artifact.value, "_emitSeq": value.get("_emitSeq", 0)}
                 self.set_parameter_value("recording", processed)
-                self.publish_update_to_parameter("recording", processed)
+                return
         return super().after_value_set(parameter, value)
 
     def _save_recording(self, recording: dict) -> VideoUrlArtifact:
--- a/griptape_nodes_library/video/video_capture.py
+++ b/griptape_nodes_library/video/video_capture.py
@@ -1,4 +1,4 @@
-import base64
+from pathlib import Path
 
 from griptape.artifacts import VideoUrlArtifact
 from griptape_nodes.exe_types.core_types import ParameterMode
@@ -6,6 +6,12 @@ from griptape_nodes.exe_types.node_types import DataNode
 from griptape_nodes.exe_types.param_components.project_file_parameter import ProjectFileParameter
 from griptape_nodes.exe_types.param_types.parameter_dict import ParameterDict
 from griptape_nodes.exe_types.param_types.parameter_video import ParameterVideo
+from griptape_nodes.files.file import FileDestination
+from griptape_nodes.retained_mode.events.project_events import (
+    AttemptMapAbsolutePathToProjectRequest,
+    AttemptMapAbsolutePathToProjectResultSuccess,
+)
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 from griptape_nodes.traits.widget import Widget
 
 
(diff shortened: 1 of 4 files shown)


griptape_nodes_library/video/video_capture.py
```diff
@@ -0,0 +1,98 @@
...
+                    server_url = GriptapeNodes.StaticFilesManager().static_server_base_url
+                    dest = self._output_file.build_file()
+                    # resolve() fails for CREATE_NEW situations because _index is only
+                    # computed at write time. Call the base-class write_bytes directly to
+                    # allocate the indexed path without the ProjectFileDestination macro
+                    # mapping — that gives us the raw absolute path we need for the URL.
+                    placeholder = FileDestination.write_bytes(dest, b"")
+                    abs_path = Path(placeholder.location)
+                    workspace = GriptapeNodes.ConfigManager().workspace_path
+                    rel_path = str(abs_path.relative_to(workspace))
```
cjkindel: AI Assisted Comment:
> *An out-of-workspace output path crashes after_value_set and strands the widget on "Processing…"*
> `abs_path.relative_to(workspace)` raises `ValueError` whenever the resolved file isn't under the workspace directory — reachable via an absolute `output_file` (ProjectFileParameter honors absolute filenames verbatim), an upstream FileOutputSettings pointing elsewhere, or a project rooted outside the workspace. When it raises, no `upload_ready` ever reaches the widget and the encoding overlay has no timeout or error state, so the artist gets an infinite spinner with re-record as the only escape. The constraint is structural — the static server's PUT route only writes workspace-relative paths — so this case should be detected and surfaced as a visible, artist-readable failure rather than an unhandled exception.

griptape_nodes_library/video/video_capture.py
```diff
@@ -0,0 +1,98 @@
...
+                    upload_url = f"{server_url}/static-uploads/{rel_path}"
```
cjkindel: AI Assisted Comment:
> *Recorded bytes bypass the engine write pipeline, so Firefox saves webm bytes into a .mp4 file*
> The zero-byte placeholder goes through OSManager (sniff, extension coercion, sidecar), but the real recording is PUT through the static server's raw `write_bytes`, which does none of that. The widget's `getSupportedMime()` falls back to webm on browsers without MediaRecorder MP4 support (Firefox), so those users get a webm container in a hard-coded `.mp4` name — exactly the mismatch `coerce_extension_to_match_bytes` exists to prevent — and the sidecar records a 0-byte write as provenance. The widget already knows the actual mime; sending it with `requesting_upload_url` and letting Python pick the extension would close this.

widgets/VideoCapture/VideoCapture.js
```diff
@@ -0,0 +1,400 @@
...
+  }
+
+  // ── handleUpdate ─────────────────────────────────────────────────────────
+
+  function handleUpdate(newProps) {
+    onChange = newProps.onChange;
+    const v = newProps.value ?? {};
+    if (v.state === "processed") { hideEncodingOverlay(); return; }
+    // Python responds to requesting_upload_url with the PUT destination and artifact path
+    if (v.state === "upload_ready" && v._uploadUrl) { doUpload(v._uploadUrl, v._artifactUrl); return; }
```
cjkindel: AI Assisted Comment:
> *upload_ready handling is not idempotent — re-renders during an in-flight upload start duplicate PUTs*
> `handleUpdate` calls `doUpload` every time it sees `state: "upload_ready"`, and the re-mount guard at the top of the factory routes every host re-render into `handleUpdate` with the still-current value. The parameter stays `upload_ready` for the whole upload (seconds for a long clip), so any canvas re-render in that window fires a second concurrent PUT of the same blob plus a second `accepted` emit. The `_emitSeq` counter threaded through both sides looks like it was built to be exactly this guard, but it's never compared anywhere. A one-flag "already uploading this seq" check fixes it.

griptape_nodes_library/video/video_capture.py
```diff
@@ -0,0 +1,98 @@
...
+        if parameter.name == "recording" and isinstance(value, dict):
+            match value.get("state"):
+                case "requesting_upload_url":
+                    server_url = GriptapeNodes.StaticFilesManager().static_server_base_url
+                    dest = self._output_file.build_file()
+                    # resolve() fails for CREATE_NEW situations because _index is only
+                    # computed at write time. Call the base-class write_bytes directly to
+                    # allocate the indexed path without the ProjectFileDestination macro
+                    # mapping — that gives us the raw absolute path we need for the URL.
+                    placeholder = FileDestination.write_bytes(dest, b"")
```
cjkindel: AI Assisted Comment:
> *Abandoned accepts leave zero-byte indexed .mp4 litter and burn _v### indices*
> Every Accept click immediately writes a zero-byte placeholder to allocate the next index. If the PUT fails, the widget silently reverts (the `.catch` in `doUpload` gives the user no feedback at all) and the empty file stays behind; each retry allocates the next index. Zero-byte litter in output directories has bitten us before (the Decode Media CREATE_NEW candidate-poisoning issue on Windows). At minimum surface the failure to the user; ideally delete the placeholder on the failure path.

widgets/VideoCapture/VideoCapture.js
```diff
@@ -0,0 +1,400 @@
...
+  acceptBtn.addEventListener("click",   accept);
+  discardBtn.addEventListener("click",  discard);
+  videoSel.addEventListener("change",   switchDevices);
+  audioSel.addEventListener("change",   switchDevices);
+  navigator.mediaDevices.addEventListener("devicechange", populateDevices);
+
+  container._vcInst = { handleUpdate, cleanup, wrapper };
+
+  const init = props.value ?? {};
+  if ((init.state === "recorded" || init.state === "accepted") && init.value) {
```
cjkindel: AI Assisted Comment:
> *restore() and its init branch can never execute*
> The init path looks for `state: "recorded"` (never produced anywhere) or `state: "accepted"` with base64 in `.value` (the accepted dict carries `url`; nothing ever stores base64 — the PR body itself says "raw bytes, no base64"). It's ~30 lines of leftover from the earlier base64 design that will mislead the next reader about what persists across reloads.

griptape_nodes_library/video/video_capture.py
shhlife: hmm - we want to be able to write to anywhere the user has defined, even if it's outside the workspace. investigating.

shhlife: okay, fixed this and now it can record to abs path or to the project

cjkindel (approved):

shhlife:
ack - just did some testing and hit an issue - fixing it real quick

PR #541, opened 2026-08-19
fix(proxy): name the License and API key when a proxy credential is missing

### LINKED ISSUE #527: Proxy node credential error names GT_CLOUD_API_KEY even when a License was checked
## Problem

When a proxy node can't resolve a credential, the error names `GT_CLOUD_API_KEY` regardless of which credentials were actually checked:

```
<Node> is missing GT_CLOUD_API_KEY. Ensure it's set in the environment/config.
```

But `resolve_proxy_api_key()` checks three sources — `GT_CLOUD_PROXY_API_KEY`, then the Griptape Nodes License (`GRIPTAPE_NODES_LICENSE`), then `GT_CLOUD_API_KEY`. The message only mentions the last one. A License-only user reading it is told to set a credential they deliberately aren't using, and given no hint that a License is accepted or that theirs wasn't found.

The message is also indistinguishable between genuinely different states: no credential at all, a License that exists but isn't visible to this process, and a `GT_CLOUD_API_KEY` registered as an empty string (which is present enough to look configured and resolves to nothing).

This comes from `_validate_api_key()` interpolating `self.API_KEY_NAME`:

```python
api_key = resolve_proxy_api_key(self.API_KEY_NAME)
if not api_key:
    msg = f"{self.name} is missing {self.API_KEY_NAME}. Ensure it's set in the environment/config."
    raise ValueError(msg)
```

`API_KEY_NAME` is the third fallback's name, not a description of what was required. The same pattern appears across the proxy nodes — `google_image_generation.py`, `qwen_image_generation.py`, `qwen_image_edit.py`, `topaz_image_enhance.py`, `wan_animate_generation.py`, `omnihuman_subject_detection.py`, `omnihuman_subject_recognition.py`, and the shared `proxy/griptape_proxy_node.py` — so a fix in one place likely covers all of them.

## Origin

Reported by Iulia Giurca via customer support. A user hit this on `Google Nano Banana Image Generation` (`gemini-3-pro-image`) while running a published Nuke gizmo. Their library was at 0.82.0, which already contains the License-aware resolver from #447, and they had an active License.

Because the error named `GT_CLOUD_API_KEY`, the natural first read was that #447 had regressed or that the library predated the fix. It took two rounds of digging into the resolver, the release history, and the version at their exact tag to establish that the resolver was correct and had simply found nothing — the credential wasn't reaching that process (tracked separately on the Nuke library side, griptape-ai/griptape-nodes-library-nuke#101). A message naming both credentials would have pointed at the real problem immediately.

## Impact

Every License-only user who hits a credential-resolution failure on a proxy node gets pointed at the wrong credential. The cost is diagnostic: the message actively misdirects, and it misdirects toward a closed issue, which makes a working fix look broken. Worst on surfaces where a credential can legitimately be missing from the process environment (published gizmos, farm jobs, any child engine with an isolated config dir), since that's exactly when the distinction between "no License" and "License not visible here" matters.

## Repro / evidence

Run any proxy node in an environment where neither `GT_CLOUD_PROXY_API_KEY`, `GRIPTAPE_NODES_LICENSE`, nor a non-empty `GT_CLOUD_API_KEY` is resolvable. The error names only `GT_CLOUD_API_KEY`.

From the reported logs, the empty-string case is real and reads confusingly — these two lines appear 7ms apart:

```
DEBUG  Secret 'GT_CLOUD_API_KEY' found in 'environment variables'
DEBUG  No GT_CLOUD_API_KEY found, skipping user fetch
```

Found, and simultaneously not found: it's registered with an empty value.

Code: `griptape_nodes_library/proxy/provider_asset_access.py` (`resolve_proxy_api_key`, resolution order documented in its docstring), and `_validate_api_key()` in the proxy nodes listed above.

## Possible direction

The engine already has a `MISSING_CREDENTIAL_MESSAGE` for this situation, which may be reusable here rather than composing something new. Worth considering whether the message can distinguish the states the resolver already knows apart — it knows whether a License was present but rejected versus absent entirely — though that's an implementation call for whoever picks this up.

collindutter:
Closes #527

### DIFF
--- a/griptape_nodes_library/image/google_image_generation.py
+++ b/griptape_nodes_library/image/google_image_generation.py
@@ -22,7 +22,6 @@ from griptape_nodes.traits.options import Options
 from griptape_nodes.utils.artifact_normalization import normalize_artifact_list
 
 from griptape_nodes_library.proxy import GriptapeProxyNode
-from griptape_nodes_library.proxy.provider_asset_access import resolve_proxy_api_key
 from griptape_nodes_library.utils.image_utils import shrink_image_to_size
 
 logger = logging.getLogger("griptape_nodes")
@@ -409,13 +408,6 @@ class GoogleImageGeneration(GriptapeProxyNode):
 
     async def _parse_result(self, result_json: dict[str, Any], _generation_id: str) -> None:
         await self._handle_response(result_json)
-
-    def _validate_api_key(self) -> str:
-        api_key = resolve_proxy_api_key(self.API_KEY_NAME)
-        if not api_key:
-            msg = f"{self.name} is missing {self.API_KEY_NAME}. Ensure it's set in the environment/config."
-            raise ValueError(msg)
-        return api_key
 
     async def _submit_request_and_process(self, params: dict[str, Any], headers: dict[str, str]) -> None:
         post_url = urljoin(self._proxy_base, f"models/{params['model']}")
--- a/griptape_nodes_library/image/qwen_image_edit.py
+++ b/griptape_nodes_library/image/qwen_image_edit.py
@@ -20,7 +20,6 @@ from griptape_nodes.traits.options import Options
 from griptape_nodes.utils.artifact_normalization import normalize_artifact_list
 
 from griptape_nodes_library.proxy import GriptapeProxyNode
-from griptape_nodes_library.proxy.provider_asset_access import resolve_proxy_api_key
 
 logger = logging.getLogger("griptape_nodes")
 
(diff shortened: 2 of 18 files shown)


mcaloud:
`CodeQL` looks unhappy with `secret_name` logging out. Probably fine to leave as-is/dismiss, but up to you if you want to change @collindutter

collindutter:
> `CodeQL` looks unhappy with `secret_name` logging out. Probably fine to leave as-is/dismiss, but up to you if you want to change @collindutter

Yeah CodeQL freaks out any time it sees the word "secret". I think it's fine to log secret _name_.

mcaloud (approved):

PR #546, opened 2026-08-20
feat: add TopazVideoUpscale node

### LINKED ISSUE #538: Generative video upscaling (Topaz Starlight Precise) isn't reachable from the standard library
## Problem

There is no way to upscale a video in Nodes. The Griptape Cloud model proxy now serves Topaz Starlight Precise 2.5 and 2.6 (`topaz-video-slp-2.5`, `topaz-video-slp-2.6`) — generative video upscaling — but nothing in this library can invoke them.

The standard library has a Topaz *image* node (`griptape_nodes_library/image/topaz_image_enhance.py`), and the `topaz` provider block in `griptape_nodes_library.json` lists nine image operations. There is no Topaz **video** entry of any kind, so this is the first video-side Topaz surface rather than a new model added to an existing node.

Desired outcome: a workflow author can bring a video into Nodes, choose a Starlight Precise model, and get an upscaled video back.

## Origin

Reported by Mike Caloud, following the cloud-side proxy work that added these two models (tracked internally). No end-user report behind it — this closes the gap between what the proxy can do and what is reachable from Nodes.

## Impact

Video upscaling is currently unavailable in Nodes by any route. The proxy's other Topaz video models are likewise unexposed, so a node added here is also the foundation for surfacing those later.

## Repro / evidence

- `find . -iname "*topaz*"` in this repo returns only `griptape_nodes_library/image/topaz_image_enhance.py` and its integration test.
- The `topaz` provider block in `griptape_nodes_library.json` contains only `gtc_topaz_*` image keys (`enhance`, `enhance_generative`, `sharpen`, `sharpen_generative`, `denoise`, `restore_generative`, `lighting`, `matting`, `tool`) — no video model keys.

## Possible direction

One option, not the decision — whoever picks this up should judge the shape:

- A new node under `griptape_nodes_library/video/` on `GriptapeProxyNode`. `griptape_nodes_library/video/ltx_video_to_video_hdr.py` is the closest analogue (video in → proxy → video out); `image/topaz_image_enhance.py` is the reference for Topaz's model/parameter-dict conventions.
- Registration is more than the node file: a `nodes[]` entry in `griptape_nodes_library.json` with a `model_usage` declaration, plus new keys under the existing `topaz` provider in the `model_catalog` block mapping to the `topaz-video-slp-2.5` / `topaz-video-slp-2.6` provider model ids.
- Request-shape notes for whoever implements it: the proxy reads the source frame count and the requested output resolution from the request body, so both need to be sent; and the model named in the request's filter list must match the routed model id or the request is rejected up front.
- Starlight is metered per frame and is substantially more expensive per second of output than Topaz's non-generative video models. Worth surfacing in the node description or tooltip so the cost isn't a surprise. Topaz publishes the frames-per-credit figures: [Starlight Precise 2.5](https://developer.topazlabs.com/video-models/starlight/starlight-precise-2.5#pricing), [Starlight Precise 2.6](https://developer.topazlabs.com/video-models/starlight/starlight-precise-2.6#pricing), [Model Pricing](https://developer.topazlabs.com/getting-started/model-pricing).

## Open question

Confirm the two models are live and enabled in the deployed proxy before the node ships — a node pointing at an unserved model id fails at request time rather than at load time.

mcaloud:
## Summary

- Adds `TopazVideoUpscale`, a new node that upscales video with Topaz Starlight Precise via the Griptape Cloud model proxy.

Closes #538

## Changes

- New `TopazVideoUpscale` node (`griptape_nodes_library/video/topaz_video_upscale.py`) supporting Starlight Precise 2.5/2.6, with a `resize_mode` UI (width / height / width and height / percentage) mirroring `ResizeVideo`'s pattern.
- Live billing-tier badges (1080p vs. 4K) on the resolution parameters, plus a hard-cap guard at Topaz's documented 3840x2160 output ceiling so the node fails fast instead of submitting a job Topaz will reject.
- Result parsing reads the video URL from Topaz's documented `download.url` response shape.
- Registers the node in `griptape_nodes_library.json` (including the two Topaz Starlight model IDs) and wires it into `proxy_api_key_providers.py`.
- Unit tests (`tests/unit/video/test_topaz_video_upscale.py`) covering resize-mode math, badge behavior, payload building, and result parsing; an integration test (`tests/integration/test_topaz_video_upscale.py`) exercising the node end-to-end via a workflow graph.

## Test plan

- [x] `uv run ruff check` — clean
- [x] `uv run pytest tests/unit/video/test_topaz_video_upscale.py -q` — 35 passed
- [x] Ran a real Topaz upscale job end-to-end and confirmed the returned video downloads and plays correctly

### DIFF
--- a/griptape_nodes_library/video/topaz_video_upscale.py
+++ b/griptape_nodes_library/video/topaz_video_upscale.py
@@ -1,0 +1,554 @@
+from __future__ import annotations
+
+import asyncio
+import logging
+import math
+from enum import StrEnum
+from typing import Any
+
+from griptape.artifacts.video_url_artifact import VideoUrlArtifact
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.param_components.artifact_url.public_artifact_url_parameter import (
+    PublicArtifactUrlParameter,
+)
+from griptape_nodes.exe_types.param_components.project_file_parameter import ProjectFileParameter
+from griptape_nodes.exe_types.param_types.parameter_dict import ParameterDict
+from griptape_nodes.exe_types.param_types.parameter_int import ParameterInt
+from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
+from griptape_nodes.exe_types.param_types.parameter_video import ParameterVideo
+from griptape_nodes.files.file import File, FileLoadError
+from griptape_nodes.traits.options import Options
+from griptape_nodes.traits.slider import Slider
+
+from griptape_nodes_library.media import coerce_media_url_or_data_uri
+from griptape_nodes_library.proxy import GriptapeProxyNode
+from griptape_nodes_library.utils.ffmpeg_utils import VideoMetadata, extract_video_metadata_structured
+
+logger = logging.getLogger("griptape_nodes")
+
+__all__ = ["TopazVideoUpscale"]
+
+MODEL_MAPPING = {
+    "Starlight Precise 2.6": "topaz-video-slp-2.6",
+    "Starlight Precise 2.5": "topaz-video-slp-2.5",
+}
+
+DEFAULT_MODEL = "Starlight Precise 2.6"
+
+# Topaz caps a Starlight job at 9000 frames. The proxy clamps the *billed* volume to
+# match, so exceeding it does not overcharge -- but Topaz still rejects the job, and
+# failing here is cheaper than discovering it after the upload.
+MAX_STARLIGHT_FRAMES = 9000
+
+# Starlight has two rate tiers, chosen by output pixel *area*, not height. A portrait
+# 1080x1920 output bills as 1080p; 1921x1080 already bills as 4K.
+# https://developer.topazlabs.com/getting-started/model-pricing
+STARLIGHT_1080P_MAX_PIXELS = 1920 * 1080
+
+# Topaz's documented hard output ceiling for Starlight (distinct from the 1080p/4K
+# billing-tier boundary above): https://docs.topazlabs.com/video-ai/project-starlight
+STARLIGHT_MAX_OUTPUT_PIXELS = 3840 * 2160
+
+COST_BADGE_MESSAGE = (
+    "Starlight is metered **per frame**, not per second, and costs far more than "
+    "Topaz's non-generative video models.\n\n"
+    "Two rate tiers, picked by output pixel area:\n"
+    "- **1080p** (up to 1920x1080) — the cheaper tier\n"
+    "- **4K** (anything larger) — roughly **2.2x** the 1080p rate\n\n"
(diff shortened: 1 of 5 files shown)


mcaloud:
I took a stab at the resize UI (width/height/percentage etc), mimicking the `Resize Video` closely. There's also text around pricing in the `model` and `resize_mode` icons. Please let me know what you think @shhlife thanks!

griptape_nodes_library/video/topaz_video_upscale.py
```diff
@@ -0,0 +1,554 @@
...
+            )
+            return
+
+        self.parameter_output_values["video_output"] = VideoUrlArtifact(value=saved.location, name=saved.name)
+        self._set_status_results(
+            was_successful=True,
+            result_details=f"Upscale successful. Video saved as {saved.name}.",
+        )
+
+    def _handle_payload_build_error(self, e: Exception) -> None:
```
cjkindel, line 543: AI Assisted Comment:
> *Validation errors no longer crash the flow when no failure edge is wired*
> This override matches the ones in topaz_image_enhance.py:1017-1024 and flux_image_generation.py:491-498 except it omits the `self._handle_failure_exception(e)` call before returning. That call is what raises when the failure output has no connection, so every ValueError guard here (missing video, underivable frame count, the 9000-frame cap, the 3840x2160 hard cap, no public URL) sets `was_successful=False` and quietly dead-ends the flow instead of crashing with immediate feedback like every other proxy node.

griptape_nodes_library/video/topaz_video_upscale.py
```diff
@@ -0,0 +1,554 @@
...
+
+        # Upload after probing: the probe needs the original local path, and there
+        # is no point paying for an upload if the file turns out to be unusable.
+        video_url = self._public_video_url_parameter.get_public_url_for_parameter()
+        if not video_url:
+            msg = f"{self.name} could not produce a public URL for the input video."
+            raise ValueError(msg)
+
+        source: dict[str, Any] = {
+            "container": "mp4",
```
cjkindel, line 465: AI Assisted Comment:
> *`container` is hardcoded to "mp4" while the node accepts any video*
> `ParameterVideo` plus the ffprobe probe happily accept a `.mov`/`.webm`/`.mkv` source: it passes every local guard, gets uploaded to cloud storage, and only then does Topaz receive a payload claiming the file is an mp4: exactly the pay-for-the-upload-then-fail path the frame-cap and pixel-cap guards were built to avoid. The probe already has what's needed (`optional_format_name`, ffmpeg_utils.py:278), so either derive `container` from it or reject non-mp4 sources up front with the other pre-upload checks.

cjkindel (approved):
Two semi-nits that seem good/easy to address but not going to block

PR #549, opened 2026-08-21
Add Topaz Astra 2 support to TopazVideoUpscale

mcaloud:
## Summary

- Folds Topaz Astra 2 into the existing `TopazVideoUpscale` node as a third model option (rather than a separate node), since it shares the same resolution UI, source probing, and upload/result plumbing as Starlight.
- Adds Astra's four creative filters (`creativity`, `prompt`, `sharp`, `realism`), its prompt-dependent frame cap (9000 unprompted / 450 prompted), and no output-resolution ceiling (Topaz publishes none for Astra).
- Extends the cost badge to state the cross-model comparison explicitly, since the within-family 4K multipliers (1.67x for Astra, 2.2x for Starlight) invite the wrong read on which model actually costs more per frame.
- Addresses two review comments left on the original TopazVideoUpscale PR (#546) by @cjkindel:
  - `_handle_payload_build_error` now calls `_handle_failure_exception`, matching `topaz_image_enhance.py` and `flux_image_generation.py`, so validation failures raise (rather than silently dead-ending the flow) when no failure edge is wired.
  - `source.container` is now derived from the video's extension (or MIME type for a data URI) instead of being hardcoded to `"mp4"`. Deriving it from ffprobe's `optional_format_name` (as originally suggested) turned out to be unreliable -- mp4/mov share one demuxer string and mkv/webm share another -- so extension-based mapping with a pre-upload rejection of unsupported containers is used instead.

## Testing

- `uv run pytest tests/unit/video/test_topaz_video_upscale.py -q` -- 70 passed
- `uv run pytest tests/unit -q` -- full unit suite passes
- `make check` -- format, ruff, pyright, and JSON manifest checks all clean
- Added an integration workflow (`tests/integration/test_topaz_video_upscale_astra.py`) exercising an Astra 2 run end-to-end, since the `filters` passthrough shape can't be verified by unit tests alone

### DIFF
--- a/griptape_nodes_library/video/topaz_video_upscale.py
+++ b/griptape_nodes_library/video/topaz_video_upscale.py
@@ -13,6 +13,7 @@ from griptape_nodes.exe_types.param_components.artifact_url.public_artifact_url_
 )
 from griptape_nodes.exe_types.param_components.project_file_parameter import ProjectFileParameter
 from griptape_nodes.exe_types.param_types.parameter_dict import ParameterDict
+from griptape_nodes.exe_types.param_types.parameter_float import ParameterFloat
 from griptape_nodes.exe_types.param_types.parameter_int import ParameterInt
 from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
 from griptape_nodes.exe_types.param_types.parameter_video import ParameterVideo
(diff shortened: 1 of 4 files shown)


griptape_nodes_library/video/topaz_video_upscale.py
```diff
@@ -350,6 +618,28 @@ def _probe_source(self, video_input: Any) -> VideoMetadata:
...
+            header = video_url.removeprefix("data:").split(",", 1)[0]
+            token = header.split(";", 1)[0].split("/", 1)[-1].lower()
+        else:
+            token = Path(urlsplit(video_url).path).suffix.lstrip(".").lower()
+
+        container = CONTAINER_BY_TOKEN.get(token)
+        if container is None:
+            found = token or "no extension"
+            msg = f"{self.name}: Topaz only accepts mp4, mov, or mkv source video, but got {found!r}."
+            raise ValueError(msg)
```
collindutter, line 525: > [!NOTE]

**Finding: extensionless video URLs now hard-fail before upload.**

Deriving `source.container` from the input is the right fix for #558, but the `container is None` branch treats "unrecognized extension" and "no extension at all" identically. A URL whose path carries no suffix (a raw storage key, a signed URL that strips the filename, a redirect target) produces `token == ""`, misses `CONTAINER_BY_TOKEN`, and raises before the job is ever submitted. Previously that same input was sent as a hardcoded `"mp4"` and at least reached Topaz.

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/3d55f3863d8a011befcadf6f408b2c62df68bb70/griptape_nodes_library/video/topaz_video_upscale.py#L86-L93

**Impact:** Low in practice. Every video-producing node in this library names its output with a `.mp4` extension, so node-to-node wiring is unaffected. The exposure is limited to externally supplied URLs, where a workflow that used to succeed on an MP4-behind-an-extensionless-URL now fails at payload build.

**Suggested fix:** Split the two cases. Keep the hard `ValueError` when a token is present but unsupported (`.webm`, `.avi` — genuinely wrong input, worth failing loudly), and for the empty-token case fall back to `"mp4"` with a `logger.warning` naming the URL. That preserves the pre-change behavior for the one input shape where the code has no signal either way, while still surfacing it. If failing closed here is deliberate, the docstring is the place to say so, since the current wording explains the ffprobe independence but not the strictness.

griptape_nodes_library/video/topaz_video_upscale.py
```diff
@@ -31,34 +34,141 @@
...
+# How the two families compare *to each other* per frame, at the same tier. Griptape
+# Cloud bills Astra 13,000 credits/frame at 1080p against Starlight's 4,992, and 21,667
+# against 10,908 at 4K -- so Astra is the pricier model by 2.6x and 2.0x respectively.
+# (griptape-cloud credits migrations 0087 and 0088.)
+#
+# This is spelled out in the badge because the RATE_SPREADS above are *within* a family
+# and invite exactly the wrong read on their own: Astra's 1.67x is the smaller number
+# but the more expensive model.
+ASTRA_VS_STARLIGHT_1080P = "2.6x"
+ASTRA_VS_STARLIGHT_4K = "2x"
```
collindutter: > [!NOTE]

**Finding: unconfirmed cross-model pricing is stated to users as settled fact.**

The arithmetic behind these constants checks out against the credit figures in the comment above (13,000/4,992 ≈ 2.6x, 21,667/10,908 ≈ 2x), and the reasoning for surfacing them — that the within-family `RATE_SPREADS` invite the wrong read on their own — is sound. The problem is confidence, not math: #558 lists these ratios under its own open questions as an assumption pending vendor/finance confirmation, but the badge renders them as flat statements ("Astra costs about **2.6x Starlight** per frame at 1080p").

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/3d55f3863d8a011befcadf6f408b2c62df68bb70/griptape_nodes_library/video/topaz_video_upscale.py#L142-L165

**Impact:** Users make a model-selection cost decision on a number the PR itself does not yet stand behind. If the confirmed ratio differs, the badge is wrong in the direction that costs money, and it is wrong in prose rather than in a value someone would think to grep for.

**Suggested fix:** Either confirm the ratios before merge and drop the open question from #558, or hedge the badge wording until they are ("roughly", "approximate — pending confirmation"). Worth a `# TODO(#558)` next to these two constants either way, so the follow-up has an anchor in the code and not only in the issue.

tests/unit/video/test_topaz_video_upscale.py
```diff
@@ -146,7 +164,9 @@ def test_odd_source_still_yields_even_output() -> None:
 
 
 def test_output_exceeding_the_hard_cap_raises() -> None:
+    # The cap is now Starlight's alone, so pin the model rather than lean on the default.
```
collindutter: > [!NOTE]

**Finding: comment describes the change rather than the code.**

"The cap is **now** Starlight's alone" is written for a reader who saw the previous version of this test. Six months from now there is no "now" and no default to lean on — just a test that pins a model. `CLAUDE.md` asks for evergreen comments that describe the code as it is, not how it evolved.

**Impact:** Cosmetic, but it compounds: temporal comments accumulate into a changelog embedded in the test file, and each one costs a future reader a detour through `git log` to decode.

**Suggested fix:**

```suggestion
    # The output ceiling belongs to Starlight only, so pin the model explicitly.
```

tests/unit/video/test_topaz_video_upscale.py
```diff
@@ -275,6 +295,21 @@ async def test_build_payload_omits_filters(monkeypatch: pytest.MonkeyPatch) -> N
     assert "filters" not in payload
 
 
+@pytest.mark.asyncio
+async def test_starlight_payload_has_only_source_and_output(monkeypatch: pytest.MonkeyPatch) -> None:
+    # The regression lock for the Astra merge: adding a second model family must not
+    # change a single byte of what Starlight sends. Asserting the exact key set (rather
+    # than "filters" not in payload) catches any new top-level key, not just that one.
```
collindutter: > [!NOTE]

**Finding: comment is framed around "the Astra merge" instead of the invariant.**

Same evergreen-comment issue as the one on line 167, and the more load-bearing of the two: "the regression lock for the Astra merge" and "adding a second model family must not change a single byte" pin the test's stated purpose to one historical event. The invariant it actually guards — Starlight's payload has exactly two top-level keys — outlives that merge and holds against any future family, not just Astra. The second half of the comment, explaining why the exact key set beats `"filters" not in payload`, is genuinely useful and worth keeping verbatim.

**Impact:** Cosmetic, with a small correctness cost: a reader who takes the comment literally may think the test is safe to delete once Astra has shipped and settled.

**Suggested fix:**

```suggestion
    # Starlight sends nothing but source and output. Asserting the exact key set
    # (rather than `"filters" not in payload`) catches any unexpected top-level key,
    # not just that one.
```

griptape_nodes_library/video/topaz_video_upscale.py
collindutter: I'm a little unsure about baking any sort of cost information at this level. Do have we done this for other nodes?

collindutter (approved):
Minor feedback on the cost badge. Nothing blocking though.

griptape_nodes_library/video/topaz_video_upscale.py
mcaloud: Good catch — fixed in 38210d9.

Split the two cases as you suggested: a recognizable but unsupported token (`.webm`, `.avi`) still raises before the upload, but an empty token now falls back to `mp4` with a `logger.warning` naming the URL. Added a `DEFAULT_CONTAINER` constant and extended the docstring to state the asymmetry explicitly — "strict about a wrong answer, lenient about no answer" — since you're right that the previous wording explained the ffprobe independence but not the strictness.

Two tests cover the new branch (extensionless URL, and a data URI with no MIME subtype); I verified both fail against the old code. `test_unsupported_container_raises_before_uploading` is unchanged — its zero-upload assertion is the part worth protecting.

griptape_nodes_library/video/topaz_video_upscale.py
mcaloud: I went and checked these against the source, and they hold up — so I've kept the numbers and instead recorded their provenance in the comment (38210d9).

The ratios come from griptape-cloud credits migrations 0087/0088, and both migrations' frames-per-credit inputs match what Topaz publishes per model:

| | fpc (1080p / 4K) | credits/frame | margin |
|---|---|---|---|
| Starlight Precise 2.5, 2.6 | 26.04 / 11.92 | 4,992 / 10,908 | 1.30x |
| Astra 2 (`ast-2`) | 10 / 6 | 13,000 / 21,667 | 1.30x |

13,000/4,992 = 2.60x and 21,667/10,908 = 1.99x, and all four rows land on a flat 1.3x margin — so the rate card is internally consistent, not a guess. The $0.10/credit figure the migrations describe as unconfirmed is in fact our contracted rate; those comments are stale.

One trap worth flagging, because it caught me for a while: the Astra **family** page publishes 7.7 / 4.3 fpc, but those are Astra 1 (`slc-1`, deprecated), not the `ast-2` this node routes. Sanity-checking 0088 against the family page makes the rate card look ~30% too cheap and the margin look near-zero. The correct source is the [Astra 2 page](https://developer.topazlabs.com/video-models/astra/astra-2). I've written that warning into the comment so the next person doesn't repeat it.

Given that, I don't think the badge needs hedging — but I'd rather you make that call than me, so say the word and I'll soften it.

I'll also update #558 to close out its open question on credit rates. Separately, #558's problem statement has this backwards — it says Astra is "the cheaper, faster option" and that Starlight is metered "at a substantially higher rate." Astra 2 is the pricier model by ~2.6x per frame. Correcting that on the issue too.

mcaloud: Yes, it's a fairly well-established pattern here — the closest analogue is Seedream:

- **`image/seedream_image_generation.py:459-477`** — `_update_size_badge()`, a conditional badge on `size` titled *"2K costs twice as much"*, whose whole purpose is communicating a relative price multiplier. Same shape as this one.
- **`video/ltx_video_extend.py:131-137`** — static `info` badge on `duration` explaining `ceil(context + duration)` billing; **`:196-210`** — conditional badge warning that billing will use the full per-request frame cap. The repo's own badge skill cites this one as a canonical example (`.claude/skills/add-parameter-badge/SKILL.md:118`, *"warns when `context=0` changes billing"*), and lists `cloud-upload` as the variant for "billing, size limits, remote storage".
- **`splat/world_labs_world_generation.py:157`** — tooltip with an explicit credit count: *"skips pano generation, saves 80 credits"*.
- **Tripo** (`three_d/tripo_{text,image,multiview}_to_3d_generation.py:128,148,159`) — *"costs more when enabled"* / *"'detailed' costs more."*
- **`three_d/_tripo_utils.py:113-117`** — appends the real charged amount to `result_details`: `f" Tripo charged {consumed} credits."`

Worth the counter-example too: `image/topaz_image_enhance.py` has no badges and no cost references at all. So it isn't universal — though image enhance is a flat per-image operation without the per-frame × tier × model-family blowup that motivates this one.

The consistent convention across all of them is **relative** cost language — multipliers, tiers, "costs more" — never dollar figures (`grep -rE '\$[0-9.]'` over `griptape_nodes_library/` returns nothing). This badge follows that: the raw credit rates stay in a source comment, and only the ratios reach the UI.

Also worth noting the per-frame/tier badges on this node aren't new here — they shipped in #546. This PR only adds the cross-model line.

mcaloud: I can also just take out the cost info @collindutter

collindutter: Since the other Topzaz node doesn't have it let's remove. We've got longer term plans to do this in a unified way.

griptape_nodes_library/video/topaz_video_upscale.py
```diff
@@ -88,27 +143,69 @@ class TopazVideoUpscale(GriptapeProxyNode):
...
+            tooltip="Topaz video model to upscale with. Astra adds creative controls; Starlight does not.",
             allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
             traits={Options(choices=list(MODEL_MAPPING.keys()))},
         )
-        model_param.set_badge(
-            variant="warning",
-            title="Billed per frame",
-            message=COST_BADGE_MESSAGE,
-        )
         self.add_parameter(model_param)
```
cjkindel, line 157: AI Assisted Comment:
> *The diff deletes the cost guidance the PR body and #558 say it adds*
> The PR body claims it "extends the cost badge to state the cross-model comparison explicitly," and #558's desired outcome is "honestly-labelled cost guidance" for Astra. The diff does the opposite: `COST_BADGE_MESSAGE` is deleted, the model parameter is built with no badge, the 1080p/4K tier badges are gone, and "Billed per frame" was removed from both the manifest description and the node description. `test_model_carries_no_badge` locks the reversal in, replacing a test whose comment said "the issue asks for that to be visible." Either the wrong revision was pushed or the body is stale for a deliberate reversal neither document records; either way this quietly regresses the per-frame cost visibility #546 shipped, on a node that bills substantially per frame.

griptape_nodes_library/video/topaz_video_upscale.py
```diff
@@ -230,104 +332,146 @@ def _get_api_model_id(self) -> str:
         model_name = self.get_parameter_value("model") or DEFAULT_MODEL
         return MODEL_MAPPING.get(model_name, MODEL_MAPPING[DEFAULT_MODEL])
```
cjkindel, line 333: AI Assisted Comment:
> *The billing path falls back silently while the cosmetic path raises*
> `_family()` raises on an unmapped model name, justified as "quietly billing at another model's rate is a worse outcome than a loud failure" -- but `_get_api_model_id`, the method that actually picks what gets billed, does exactly that quiet thing: `MODEL_MAPPING.get(model_name, MODEL_MAPPING[DEFAULT_MODEL])` silently routes a stale name to Starlight 2.6. The Options trait only protects live UI input; a workflow saved before a future model rename/removal replays the old string, at which point routing silently swaps the billed model while every badge/visibility reaction touching `_family()` (fired from `after_value_set` on `model`, `resize_mode`, and all four size parameters) raises `ValueError` from UI plumbing. One value, two contradictory policies, with the loud one guarding the harmless path. `_get_api_model_id` should raise on a miss too, or both should share one lookup.

griptape_nodes_library/video/topaz_video_upscale.py
```diff
@@ -350,6 +494,44 @@ def _probe_source(self, video_input: Any) -> VideoMetadata:
...
+            msg = f"{self.name}: Topaz only accepts mp4, mov, or mkv source video, but got {token!r}."
+            raise ValueError(msg)
+
+        logger.warning(
+            "%s could not determine a container for %s (no extension or MIME subtype); assuming %s.",
+            self.name,
+            video_url,
+            DEFAULT_CONTAINER,
+        )
+        return DEFAULT_CONTAINER
```
cjkindel, line 533: AI Assisted Comment:
> *Extensionless URLs get a guessed mp4 container on a metered job*
> `_derive_container` assumes `mp4` when the input carries no extension or MIME subtype, with only a log warning. If such a URL actually serves mov/mkv bytes, Topaz gets a wrong container declaration and the job fails (or misbehaves) after the upload -- the exact late-failure cost the rest of this PR spends effort avoiding. The docstring argues the case thoughtfully (rejecting would break signed URLs Topaz would accept, and it matches the old hardcoded behavior for that input shape), so this is a judgment call rather than a defect -- flagged because it is a fallback that guesses on behalf of a metered job.

cjkindel (approved):
Only AI comments are nits, nothing blocking

PR #551, opened 2026-08-22
feat: add explicit output resolution to TopazImageEnhance

mcaloud:
Part 1 of griptape-ai/griptape-cloud#2178. The billing half is griptape-ai/griptape-cloud#2209 — **that one must merge and its migration must apply first**, or Topaz image jobs bill 0.

## What

Topaz's enhance models *are* the upscalers, but `TopazImageEnhance` had no way to say how big the output should be — it always took whatever Autopilot picked. You couldn't target 4K, hit a print size, or do a plain 2x.

Topaz accepts three top-level **snake_case** fields — `output_width`, `output_height` (1–32000) and `crop_to_fill` — on exactly three endpoints: `/enhance/async`, `/enhance-gen/async`, `/tool/async`. There is no `upscale_factor`. This PR exposes them behind a `Resize` parameter group.

| `resize_mode` | emits |
|---|---|
| `none` (default) | nothing — today's payload, byte for byte |
| `width` | `output_width` + height derived from the source aspect ratio |
| `height` | `output_height` + width derived from the source aspect ratio |
| `width and height` | both, verbatim; `crop_to_fill` also shown |
| `percentage` | both, scaled from the source dimensions |

A badge on `resize_mode` shows the estimated Topaz credit cost using the same `max(1, ceil(output_MP / rate))` meter the cloud bills on.

## Backward compatibility

Non-breaking, and there's a test that pins it. Two things make it so:

- **`output_width`/`output_height` default to `None`, not `0`.** `_build_payload` skips a parameter only when its value is `None` — a `0` default would send `output_width=0`, violate Topaz's 1–32000 range, and fail *every existing workflow on this node*.
- **Nothing depends on a load-time callback.** Workflow load runs with `initial_setup=True`, which skips all before/after hooks, so `after_value_set` never fires. `__init__` declares the group already correct for the default `enhance` / `none` state.

Workflows serialize only parameters that have a value, so a parameter that didn't exist gets no `SetParameterValueRequest` and falls back to its default. Connections are keyed by parameter name; there is no node signature or parameter-set hash anywhere in the engine.

## Deviation from the plan

The plan had `width`-only and `height`-only send a single dimension and let Topaz scale the other. The cloud biller requires **both dimensions or neither** — a lone dimension makes output megapixels unknowable at request time, which is the exact hole the billing change closes — so those modes would have 400'd. The node now derives the partner dimension from the source aspect ratio instead. The cost is that `width`/`height` need a readable source image, like `percentage` does; `width and height` is the escape hatch when the source can't be read.

## Testing

`make check` clean. `make test/unit` → **920 passed, 11 skipped**, including 51 new tests in `tests/unit/image/test_topaz_image_enhance.py`:

- default payload contains none of `output_width`/`output_height`/`crop_to_fill` (the backcompat regression test)
- `resize_mode=none` emits nothing even with stale dimensions left in the fields
- each mode's payload; width/height derive the partner dimension; every non-`none` mode emits both
- source-unreadable raises for `width`/`height`/`percentage`; `width and height` doesn't need the source
- `crop_to_fill` only ever accompanies a dimension
- `denoise`/`sharpen`/`lighting`/`matting` never emit resize keys
- validation: out of range, missing dimension, resize on a non-resizable operation, bools rejected as dimensions
- the published Topaz credit tables, row for row

## Not done

**No live call to Topaz has been made.** Topaz *silently ignores* fields it doesn't recognize rather than rejecting them, so a snake_case/camelCase mistake in `output_width` would look exactly like success — a plausible image at the wrong size. A real 2x and a real fixed-4K run, asserting the returned image's actual dimensions, is the one check that can't be replaced by unit tests. Worth doing before this leaves draft.

Also pre-existing, noted but not fixed: `webp` is in the node's `OUTPUT_FORMAT_OPTIONS` but is not in Topaz's documented output enum.

### DIFF
--- a/griptape_nodes_library/image/topaz_image_enhance.py
+++ b/griptape_nodes_library/image/topaz_image_enhance.py
@@ -2,12 +2,14 @@ from __future__ import annotations
 
 import json
 import logging
+import math
 from contextlib import suppress
 from copy import deepcopy
+from enum import StrEnum
 from typing import Any
 
 from griptape.artifacts import ImageArtifact, ImageUrlArtifact
-from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.core_types import Parameter, ParameterGroup, ParameterMode
 from griptape_nodes.exe_types.param_components.project_file_parameter import ProjectFileParameter
 from griptape_nodes.exe_types.param_types.parameter_bool import ParameterBool
 from griptape_nodes.exe_types.param_types.parameter_dict import ParameterDict
@@ -21,6 +23,7 @@ from griptape_nodes.utils.artifact_normalization import normalize_artifact_input
 
 from griptape_nodes_library.proxy import GriptapeProxyNode
 from griptape_nodes_library.proxy.provider_asset_access import resolve_proxy_api_key
+from griptape_nodes_library.utils.image_utils import get_image_dimensions_from_artifact
 
 logger = logging.getLogger("griptape_nodes")
 
@@ -41,6 +44,44 @@ OPERATION_OPTIONS = [
     "matting",
     "tool",
 ]
+
+
+class ResizeMode(StrEnum):
+    """How the requested output resolution is derived."""
+
+    NONE = "none"
+    WIDTH = "width"
+    HEIGHT = "height"
+    WIDTH_HEIGHT = "width and height"
+    PERCENTAGE = "percentage"
+
+
+# Only these three Topaz endpoints accept output_width/output_height/crop_to_fill.
+# sharpen, sharpen-generative, denoise, restore-generative, lighting and matting
+# always emit at the input resolution.
+UPSCALE_OPERATIONS = frozenset({"enhance", "enhance-generative", "tool"})
+
+# Topaz's documented bounds for a single output dimension.
+MIN_OUTPUT_DIMENSION = 1
+MAX_OUTPUT_DIMENSION = 32_000
+
+# Topaz meters output megapixels against credits, and the rate is fixed per endpoint
+# rather than per model: everything on /enhance-gen/async is Wonder-family, everything
+# on /enhance/async and /tool/async is Gigapixel-family.
+# https://developer.topazlabs.com/getting-started/model-pricing.md
+MP_PER_CREDIT_BY_OPERATION = {
+    "enhance": 24,
+    "enhance-generative": 4,
+    "tool": 24,
+}
+
+# Parameters in the resize group. Deliberately NOT part of any per-model table (and so
+# not in ALL_MODEL_PARAMS): resize is a property of the operation, not of the model, and
+# _update_visible_params_for_model hides every ALL_MODEL_PARAMS member the selected model
+# does not declare.
+RESIZE_PARAMS = ("output_width", "output_height", "percentage", "crop_to_fill")
+
+PERCENT_SCALE = 100
 
 ENHANCE_MODELS = {
     "Standard V2": [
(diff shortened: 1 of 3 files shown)


griptape_nodes_library/image/topaz_image_enhance.py
```diff
@@ -885,6 +1091,18 @@ async def _build_payload(self) -> dict[str, Any]:
...
+        # silently ignores unknown keys -- so a wrong name looks like success. They are
+        # snake_case at the top level even though the per-model settings are camelCase.
+        if operation in UPSCALE_OPERATIONS:
+            width, height = self._resolve_output_size(params)
+            if width is not None:
+                payload["output_width"] = width
+            if height is not None:
+                payload["output_height"] = height
+            if width is not None or height is not None:
+                payload["crop_to_fill"] = bool(params.get("crop_to_fill"))
```
collindutter: > [!NOTE]

**Finding: `crop_to_fill` is emitted in modes where it cannot apply.**

The key is written whenever any dimension resolves, so `WIDTH`, `HEIGHT`, and `PERCENTAGE` all carry it. Those modes derive the partner dimension from the source aspect ratio, so the requested ratio always equals the source ratio, which is the condition this code already documents as the one case where `crop_to_fill` matters:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/4141b16fc4655f443200f48461e6aeb56a0c652a/griptape_nodes_library/image/topaz_image_enhance.py#L930-L933

`crop_to_fill` is also hidden outside `WIDTH_HEIGHT`, so a value set there survives a switch to `WIDTH` and lands in the payload with no way for the user to see or clear it.

**Impact:** No functional change, Topaz has nothing to crop when the ratios match. It is payload noise that misrepresents intent in request logs, and it makes the emitted payload depend on parameter state the UI has hidden.

**Suggested fix:** Gate the key on the one mode that can use it.

```suggestion
            if params.get("resize_mode") == ResizeMode.WIDTH_HEIGHT:
                payload["crop_to_fill"] = bool(params.get("crop_to_fill"))
```

Verified: `tests/unit/image/test_topaz_image_enhance.py` passes (51 passed) and `ruff check` is clean with this applied.

griptape_nodes_library/image/topaz_image_enhance.py
```diff
@@ -894,6 +1112,128 @@ async def _build_payload(self) -> dict[str, Any]:
...
+            case ResizeMode.WIDTH:
+                width = self._validated_dimension(params.get("output_width"), "output_width")
+                source_width, source_height = self._source_dimensions(params)
+                height = max(1, round(width * source_height / source_width))
+                return (width, self._validated_dimension(height, "output_height"))
+            case ResizeMode.HEIGHT:
+                height = self._validated_dimension(params.get("output_height"), "output_height")
+                source_width, source_height = self._source_dimensions(params)
+                width = max(1, round(height * source_width / source_height))
+                return (self._validated_dimension(width, "output_width"), height)
```
collindutter: > [!NOTE]

**Finding: the `WIDTH` and `HEIGHT` branches are mirror images.**

Both do the same four steps: validate the known dimension, read the source dimensions, derive the partner from the source ratio, validate the derived value. Only the axis differs, and the axis swap is spread across three lines in each branch rather than expressed once.

**Impact:** Minor. The duplication is small and correct today, but the ratio math is the part that keeps proxy billing exact, so a future change to rounding or clamping has to be made identically in two places to stay consistent.

**Suggested fix:** Optional, not blocking. A single helper taking the known value, its parameter name, and the partner's name, and deriving the partner from the source ratio, collapses both branches to one call each and leaves one copy of the math.

collindutter (approved):

collindutter (approved):

PR #552, opened 2026-08-22
Add TopazVideoConvertHdr for SDR-to-HDR conversion via Topaz Hyperion

mcaloud:
Adds `TopazVideoConvertHdr`, a node that converts an SDR video to HDR with Topaz Hyperion through the Griptape Cloud model proxy.

Node half of griptape-ai/griptape-cloud#2205. **Blocked on griptape-ai/griptape-cloud#2208** — the node 400s at client construction until the proxy knows the model ids.

> **Base branch is `feature/topaz-astra-2`, not `main`.** That branch is still in review, rewrites 456 lines of `topaz_video_upscale.py`, and adds the `gtc_topaz_video_ast_2` catalog entry; branching from `main` conflicts in both files. Rebase once Astra merges.

## Why a new node rather than a mode on `TopazVideoUpscale`

There is no SDR-to-HDR conversion anywhere in the library today. The only HDR path is `LTXVideoToVideoHDR` — a generative model on a different provider that emits a ZIP of per-frame EXRs rather than a playable file.

Hyperion is a tone-mapper. It takes no creative controls, does not resize, and tags BT.2020/PQ itself, so it shares almost nothing with the upscale UI: no resize modes, no prompt, no tier badge branches.

## What it exposes — and what it deliberately does not

| | |
|---|---|
| `model` | Hyperion 2.5 (default) / Hyperion 2 |
| `video` | source SDR clip, uploaded and handed to Topaz as `source.external` |
| `output_format` | H.265 Main10 (mp4), default · ProRes 422 HQ (mov) |

**No transfer-function control.** Hyperion 1 had `transferFunction` (`pq`/`hlg`); 2 and 2.5 dropped it, and the schema has no color field of any kind. The model tags the output itself, so there is nothing for the node to offer and nothing for it to promise beyond what `ffprobe` shows.

**No resize.** Hyperion converts rather than scales, so `output.resolution` echoes the probed source. Topaz requires the field regardless, and the proxy reads it to pick the billing tier.

`output_format` is the one real choice, and it fans out to three request fields — `videoEncoder`, `videoProfile`, `container` — through a single table. A ProRes stream in an mp4 is not a thing, and three separate conditionals would eventually produce one.

## Both model ids

`hyp-2` and `hyp-2.5` are the same price, the same request shape, and the same activity types. `hyp-2.5` is absent from Topaz's own credit calculator and their launch post says "coming soon to Topaz Video", so it is not yet certain it is live on the REST API. Carrying both costs two catalog entries and covers either answer — and satisfies griptape-ai/griptape-cloud#1970, which asked for Hyperion 2 specifically.

## Advisories, not gates

The probe flags two cases and lets the run proceed:

- source already reports `smpte2084` / `bt2020` — converting it bills every frame for a no-op
- source reports interlaced field order — Topaz notes SDR→HDR "does not operate well with interlaced footage"

Both warn rather than block. A file mislabelled BT.2020 is real, and so is deliberately re-running a conversion; refusing either would be worse than saying so.

## ProRes is an assumption

`videoEncoder`'s published enum is `[AV1, H264, H265, VP9]` and omits ProRes — while two sibling fields in the *same* schema document it: `videoProfile` lists the four 422 profiles, and the resolution table gives ProRes a 16386 ceiling. The enum is assumed stale, the same way `source.external.provider`'s was (its real list `r2, s3, web-url` was only recoverable from the validator's error string).

**Not verified against the live API.** If Topaz rejects it, the fix is deleting one entry from `OUTPUT_ENCODINGS`.

## Shared module

`topaz_video_common.py` holds the source plumbing `TopazVideoUpscale` already had — `probe_source`, `derive_container`, `frame_count`, `to_even`, `CONTAINER_BY_TOKEN`, `TIER_1080P_MAX_PIXELS` — following the `seedance_common` precedent (module-level pure functions, no mixin). The upscale node keeps thin delegating methods, so its behaviour and its 70 existing tests are unchanged.

The neighbouring modules were checked and are all wrong homes: `public_video_url_mixin.py` is LTX-specific (its reason for existing is a base64 fallback tier Topaz never uses), `base_video_processor.py` is a local ffmpeg base with no proxy involvement, and `media/` is for genuinely provider-agnostic helpers.

## Also

`TopazVideoConvertHdr` is registered in `_NODE_PROVIDER_CONFIGS`. Skipping that **silently** drops the BYOK parameters — no error, just a node that can only ever use the Griptape key.

The result details say plainly that an HDR preview may look wrong while the file is correct. A BT.2020 PQ file is exactly what griptape-ai/griptape-vsl-gui#2724 renders too dark, and a ProRes `.mov` will not preview in a browser at all.

## Cost

Hyperion bills **per frame**, in two tiers by output pixel area (a portrait 1080x1920 bills as 1080p), at roughly **2× Starlight** — one of the more expensive video models here. The frame cap is 9,000, matching what the proxy clamps billing to; Topaz publishes no Hyperion-specific limit, so that is our conservative choice, and it fails in `_build_payload` before the upload rather than after.

## Testing

- `tests/unit/video/test_topaz_video_convert_hdr.py` — 39 tests: payload shape, format→encoder/profile/container mapping, output resolution and the per-encoder size ceiling, frame-count derivation and cap, "rejected input is never uploaded", the two advisories, filename tracking the container (and never clobbering a user-typed one), and the `_parse_result` binary-vs-`download.url` split
- `tests/unit` — 945 passed, 11 skipped
- `make check` — clean

Not yet run end to end. The acceptance criterion is `ffprobe -show_streams` on the output reporting `color_transfer=smpte2084` and `color_primaries=bt2020` — that is what distinguishes real HDR from a wide container holding SDR content, and it needs a live key and a paid job.

### DIFF
--- a/griptape_nodes_library/video/topaz_video_convert_hdr.py
+++ b/griptape_nodes_library/video/topaz_video_convert_hdr.py
@@ -1,0 +1,582 @@
+from __future__ import annotations
+
+import asyncio
+import logging
+from dataclasses import dataclass
+from enum import StrEnum
+from typing import Any
+
+from griptape.artifacts.video_url_artifact import VideoUrlArtifact
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.param_components.artifact_url.public_artifact_url_parameter import (
+    PublicArtifactUrlParameter,
+)
+from griptape_nodes.exe_types.param_components.project_file_parameter import ProjectFileParameter
+from griptape_nodes.exe_types.param_types.parameter_dict import ParameterDict
+from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
+from griptape_nodes.exe_types.param_types.parameter_video import ParameterVideo
+from griptape_nodes.traits.options import Options
+
+from griptape_nodes_library.proxy import GriptapeProxyNode
+from griptape_nodes_library.utils.ffmpeg_utils import VideoMetadata
+from griptape_nodes_library.video.topaz_video_common import (
+    derive_container,
+    frame_count,
+    probe_source,
+    to_even,
+)
+
+logger = logging.getLogger("griptape_nodes")
+
+__all__ = ["TopazVideoConvertHdr"]
+
+# Both versions are registered in the proxy at the same price and the same request
+# shape. 2.5 is the newer model and the default; 2 is kept selectable because 2.5 is
+# absent from Topaz's own credit calculator and may not yet be live on the REST API.
+MODEL_MAPPING = {
+    "Hyperion 2.5": "topaz-video-hyp-2.5",
+    "Hyperion 2": "topaz-video-hyp-2",
+}
+
+DEFAULT_MODEL = "Hyperion 2.5"
+
+# Topaz documents no Hyperion-specific frame cap, unlike Astra's explicit 9000/450.
+# This is the platform-wide job cap, which is also what the proxy clamps billing to.
+# Our conservative choice rather than a published Hyperion limit.
+MAX_HYPERION_FRAMES = 9000
+
+
+class OutputFormat(StrEnum):
+    """How the HDR result is encoded.
+
+    Values are the display spelling so they interpolate straight into badge prose and
+    read correctly in the dropdown.
+    """
+
+    H265_MAIN10 = "H.265 Main10 (mp4)"
+    PRORES_422_HQ = "ProRes 422 HQ (mov)"
+
+
+@dataclass(frozen=True)
+class OutputEncoding:
+    """The three ``output`` fields one format choice fans out to, plus its size ceiling.
+
+    Kept as one table rather than three conditionals: the encoder, the profile and the
+    container have to agree -- a ProRes stream in an mp4 is not a thing -- and splitting
+    them invites exactly that mismatch.
+    """
+
+    video_encoder: str
+    video_profile: str
+    container: str
+    # Topaz's per-encoder maximum output dimension. Hyperion does not scale, so this
+    # only bites on an unusually large source, but H.265 tops out well below ProRes.
+    # https://developer.topazlabs.com/reference/api-endpoints/video/create-request
+    max_dimension: int
+    file_extension: str
+
+
+OUTPUT_ENCODINGS: dict[OutputFormat, OutputEncoding] = {
+    # "Main10" is the 10-bit H.265 profile. 8-bit "Main" is not offered: banding in
+    # smooth gradients is the classic 8-bit HDR failure, and it would waste the
+    # conversion.
+    OutputFormat.H265_MAIN10: OutputEncoding(
+        video_encoder="H265",
+        video_profile="Main10",
+        container="mp4",
+        max_dimension=8192,
+        file_extension="mp4",
+    ),
+    # ProRes is the mastering path Topaz's own Hyperion page advertises. Note that
+    # `videoEncoder`'s published enum is [AV1, H264, H265, VP9] and omits ProRes, while
+    # two sibling fields in the *same* schema document it -- `videoProfile` lists the
+    # four 422 profiles and the resolution table gives ProRes a 16386 ceiling. The enum
+    # is assumed stale, the same way `source.external.provider`'s was. Unverified
+    # against the live API; if Topaz rejects it, drop this entry.
+    OutputFormat.PRORES_422_HQ: OutputEncoding(
+        video_encoder="ProRes",
+        video_profile="422 HQ",
+        container="mov",
+        max_dimension=16386,
+        file_extension="mov",
+    ),
+}
+
+DEFAULT_OUTPUT_FORMAT = OutputFormat.H265_MAIN10
+
+
+def _default_filename(output_format: OutputFormat) -> str:
+    """Default output filename for a format, so the extension tracks the container."""
+    return f"topaz_video_convert_hdr.{OUTPUT_ENCODINGS[output_format].file_extension}"
+
+
+# ffprobe's spellings for transfer characteristics and primaries that mean the source
+# is already HDR. Hyperion would still run -- and still bill -- but it would be
+# tone-mapping something that does not need it.
+HDR_TRANSFERS = frozenset({"smpte2084", "arib-std-b67", "smpte428", "bt2020-10", "bt2020-12"})
+HDR_PRIMARIES = frozenset({"bt2020"})
+
+# ffprobe reports progressive footage as "progressive"; anything else here is some
+# flavour of interlacing. Topaz notes that SDR-to-HDR "does not operate well with
+# interlaced footage".
+PROGRESSIVE_FIELD_ORDER = "progressive"
+
+COST_BADGE_MESSAGE = (
+    "Hyperion is metered **per frame**, not per second.\n\n"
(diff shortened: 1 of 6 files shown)


griptape_nodes_library/video/topaz_video_convert_hdr.py
```diff
@@ -0,0 +1,582 @@
...
+
+def _default_filename(output_format: OutputFormat) -> str:
+    """Default output filename for a format, so the extension tracks the container."""
+    return f"topaz_video_convert_hdr.{OUTPUT_ENCODINGS[output_format].file_extension}"
+
+
+# ffprobe's spellings for transfer characteristics and primaries that mean the source
+# is already HDR. Hyperion would still run -- and still bill -- but it would be
+# tone-mapping something that does not need it.
+HDR_TRANSFERS = frozenset({"smpte2084", "arib-std-b67", "smpte428", "bt2020-10", "bt2020-12"})
```
collindutter: > [!NOTE]

**Finding: `bt2020-10` and `bt2020-12` are not HDR transfer characteristics.**

These are ffprobe's spellings for ITU-T H.273 transfer characteristic values 14 and 15 ("BT.2020 for 10-bit system" / "12-bit system"). Both use the **same OETF as BT.709** — the spec exists to label bit depth on wide-gamut content, not dynamic range. Only `smpte2084` (PQ) and `arib-std-b67` (HLG) actually indicate HDR.

**Impact:** 10/12-bit SDR footage in a BT.2020 container (common on UHD masters) trips the "the source already looks like HDR" advisory. That is a different case from the "file mislabelled BT.2020" scenario the PR description defends for `HDR_PRIMARIES`: here the file is correctly labelled and genuinely SDR, so the advisory is simply wrong rather than defensively noisy. Advisory-only, so nothing is blocked or misbilled.

**Suggested fix:**

```suggestion
HDR_TRANSFERS = frozenset({"smpte2084", "arib-std-b67", "smpte428"})
```

griptape_nodes_library/video/topaz_video_convert_hdr.py
```diff
@@ -0,0 +1,582 @@
...
+        primaries = (color.optional_color_primaries or "").lower()
+        if transfer in HDR_TRANSFERS or primaries in HDR_PRIMARIES:
+            advisories.append(
+                f"The source already looks like HDR (transfer {transfer or 'unset'!r}, "
+                f"primaries {primaries or 'unset'!r}). Hyperion converts SDR footage; "
+                "running it on HDR input still bills for every frame."
+            )
+
+        field_order = (color.optional_field_order or "").lower()
+        if field_order and field_order != PROGRESSIVE_FIELD_ORDER:
```
collindutter: > [!NOTE]

**Finding: `field_order == "unknown"` is treated as interlaced.**

`ffprobe` emits `field_order: "unknown"` when it *could not determine* the value, not when it detected interlacing. `optional_field_order` is passed straight through from the raw stream dict with no filtering:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/ed88f9b95f7bc5d8e276b436053199118a6ad5a6/griptape_nodes_library/utils/ffmpeg_utils.py#L303

So the guard's premise — that anything other than `"progressive"` is a flavour of interlacing — does not hold:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/ed88f9b95f7bc5d8e276b436053199118a6ad5a6/griptape_nodes_library/video/topaz_video_convert_hdr.py#L119-L122

**Impact:** an ordinary progressive file that simply lacks field-order metadata produces the self-contradictory advisory `The source reports interlaced field order 'unknown'`, and tells the user to deinterlace footage that is not interlaced. Not covered by tests — `test_interlaced_footage_is_flagged` only exercises `"tt"`, and `test_progressive_footage_is_not_flagged` only `"progressive"`, so the gap sits between the two.

**Suggested fix:** exclude the sentinel. Worth updating the comment above `PROGRESSIVE_FIELD_ORDER` (linked above) to name it, and adding a `field_order="unknown"` case to the advisory tests.

```suggestion
        if field_order and field_order not in {PROGRESSIVE_FIELD_ORDER, "unknown"}:
```

griptape_nodes_library/video/topaz_video_upscale.py
```diff
@@ -19,13 +17,18 @@
...
 from griptape_nodes.traits.slider import Slider
 
-from griptape_nodes_library.media import coerce_media_url_or_data_uri
 from griptape_nodes_library.proxy import GriptapeProxyNode
-from griptape_nodes_library.utils.ffmpeg_utils import VideoMetadata, extract_video_metadata_structured
+from griptape_nodes_library.utils.ffmpeg_utils import VideoMetadata
+from griptape_nodes_library.video.topaz_video_common import (
+    TIER_1080P_MAX_PIXELS,
+    derive_container,
+    frame_count,
```
collindutter, line 26: > [!NOTE]

**Finding: this import is shadowed by a local variable in `_build_payload`.**

Hoisting `frame_count` to a module-level name collides with an existing local in `_build_payload`, which the extraction did not rename:

https://github.com/griptape-ai/griptape-nodes-library-standard/blob/ed88f9b95f7bc5d8e276b436053199118a6ad5a6/griptape_nodes_library/video/topaz_video_upscale.py#L666-L670

Before this diff no top-level `frame_count` name existed in the module, so the collision is new. The sibling new file avoids it by naming its local `source_frames`.

**Impact:** no behavioural change — `self._frame_count` still resolves, and the thin wrapper calls the module function from its own scope. But for the rest of `_build_payload` the imported function is unreachable under its own name, and a reader has to work out which `frame_count` is in play at each line.

**Suggested fix:** rename the local to `source_frame_count` (or `source_frames`, matching `topaz_video_convert_hdr`) and update its uses in `_build_payload`.

griptape_nodes_library/video/topaz_video_convert_hdr.py
```diff
@@ -0,0 +1,582 @@
...
+                "resolution": {"width": output_width, "height": output_height},
+                "videoEncoder": encoding.video_encoder,
+                "videoProfile": encoding.video_profile,
+                "container": encoding.container,
+            },
+        }
+        # No `filters`: Hyperion accepts no creative controls, and the proxy stamps the
+        # routed model code onto the filters it synthesizes itself. Sending our own
+        # would only risk the "filters[].model must match the requested model"
+        # rejection.
```
collindutter: > [!NOTE]

**Finding: this comment sits after the `return`, where it is unreachable in reading order.**

The `# No \`filters\`` explanation is placed below the `return` statement it explains. `TopazVideoUpscale` puts the equivalent comment *before* its return, so this also breaks the convention the sibling node set.

**Impact:** readability only. A reader scanning top-down reaches the comment after the function has already returned, and the "why there is no `filters` key" rationale is easy to miss where it matters — while reading the dict literal.

**Suggested fix:**

```suggestion
        # No `filters`: Hyperion accepts no creative controls, and the proxy stamps the
        # routed model code onto the filters it synthesizes itself. Sending our own
        # would only risk the "filters[].model must match the requested model"
        # rejection.
        return {
            "source": source,
            "output": {
                "resolution": {"width": output_width, "height": output_height},
                "videoEncoder": encoding.video_encoder,
                "videoProfile": encoding.video_profile,
                "container": encoding.container,
            },
        }
```

collindutter (commented):
Minor

